PR #7, opened 2015-08-14
Rename basicClassCreator.js to BasicClassCreator.js

marcelpanse:
The import is referenced with a capital from the components, but the filename is lowercased. The works on osx, but fails on our linux build-server


velveteer:
The perpetrators: 

https://github.com/tleunen/react-mdl/blob/master/src/card/Card.js#L4
https://github.com/tleunen/react-mdl/blob/master/src/layout/Layout.js#L4
https://github.com/tleunen/react-mdl/blob/master/src/Menu.js#L4

tleunen:
Thanks.
I just pushed a new version with a similar fix. I went for renaming the import instead of the filename.

btw, I think we can change the casing of a file by using `git mv` instead of making 2 commits ;)

marcelpanse:
I think you are right about git mv, but I had similar problems in the past where i didn't pick up on other machines, especially renaming folders..

Anyway, thanks for the update :-)

PR #10, opened 2015-08-22
README should instruct material .js/.css is inside the extra folder

jlc467:


tleunen:
Could you also update the paths and clean up the commit log? The merge commit is not needed

tleunen:
I manually merged it and added the missing part. Thanks!

PR #11, opened 2015-08-24
Fix Navigation

Graf009:
The component [HeaderRow] children forced to wrapped to the [Navigation], it is bad because I can not correctly embed [Button], my solution replace {this._renderNav ()} to {this.props.children}

### DIFF
--- a/src/layout/HeaderRow.js
+++ b/src/layout/HeaderRow.js
@@ -1,22 +1,10 @@
 import React, { PropTypes } from 'react';
 import classNames from 'classnames';
-import Navigation from './Navigation';
 
 class HeaderRow extends React.Component {
     static propTypes = {
         className: PropTypes.string,
         title: PropTypes.string
-    }
-
-    _renderNav() {
-        var hasChildren = React.Children.count(this.props.children);
-        if(!hasChildren) return;
-
-        return (
-            <Navigation>
-                {this.props.children}
-            </Navigation>
-        );
     }
 
     render() {
@@ -28,7 +16,7 @@ class HeaderRow extends React.Component {
             <div className={classes} {...otherProps}>
                 {title ? <span className="mdl-layout-title">{title}</span> : null}
                 <div className="mdl-layout-spacer"></div>
-                {this._renderNav()}
+                {this.props.children}
             </div>
         );
     }
--- a/src/layout/Layout.js
+++ b/src/layout/Layout.js
@@ -32,5 +32,6 @@ export default mdlUpgrade(Layout);
 export Header from './Header';
 export Drawer from './Drawer';
 export HeaderRow from './HeaderRow';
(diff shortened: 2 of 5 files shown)


Graf009:
Now I make an example http://www.getmdl.io/templates/dashboard/index.html on react-mdl

Graf009:
I'm still working on it

Graf009:
I finished

src/layout/HeaderRow.js
```diff
@@ -1,22 +1,11 @@
...
-        var hasChildren = React.Children.count(this.props.children);
-        if(!hasChildren) return;
-
-        return (
-            <Navigation>
-                {this.props.children}
-            </Navigation>
-        );
+        title: PropTypes.string,
+        children: PropTypes.object
```
tleunen: `children` not needed, you can safely remove it :)

src/layout/Layout.js
```diff
@@ -32,5 +32,6 @@ export default mdlUpgrade(Layout);
 export Header from './Header';
 export Drawer from './Drawer';
 export HeaderRow from './HeaderRow';
+export Navigations from './Navigation';
```
tleunen: Should be `Navigation` (without 's')

tleunen:
Thanks @Graf009. I added a few comments.
I think it's a good idea to put the Navigation as a separate component so the end user can do what he wants inside the header/drawer. 

Could you please read my comments and then merge all your commit into one? (`git rebase -i` should help you)

Graf009:
Ready

src/layout/HeaderRow.js
```diff
@@ -28,7 +16,7 @@ class HeaderRow extends React.Component {
             <div className={classes} {...otherProps}>
                 {title ? <span className="mdl-layout-title">{title}</span> : null}
                 <div className="mdl-layout-spacer"></div>
```
tleunen: Should we replace this with the new `<Spacer />` or remove it?
I don't see any use-cases where we would not put a spacer between the title and the children/navigation...

src/layout/Navigation.js
```diff
@@ -1,11 +1,19 @@
-import React from 'react';
+import React, { PropTypes } from 'react';
 import classNames from 'classnames';
 import cloneChildren from '../utils/cloneChildren';
 
 class Navigation extends React.Component {
+	static propTypes = {
+        className: PropTypes.string
```
tleunen: indentation is not right here

tleunen:
Thanks so much! Please rebase, fix the indentation and then I'll merge :)

Graf009:
I can not do rebase can help me?

tleunen:
Done, thanks!
There are a few other things I wanna add to the project before releasing a new version. Hope that's ok

PR #12, opened 2015-08-24
Add Spacer component

Graf009:
Finish


tleunen:
Can we have a spacer inside a navigation? Do you have an example?

Graf009:
For Example http://www.getmdl.io/templates/dashboard/index.html: drawer has a main menu, and in the bottom of the help button

tleunen:
Thanks @Graf009. I manually merged it because I changed a few details from your implementation ;-)

PR #19, opened 2015-08-27
Update package.json

web2style:


tleunen:
No need to create a PR from your master to this master.
You can create it directly from your `patch-1` branch to this master. This way, only 1 commit will show up (with your package.json update), without your merge commit.

Please resubmit the PR and it will be good :)

PR #22, opened 2015-08-28
Fixed error if there are no children in the header

Graf009:
Fixed error if there are no children in the header

### DIFF
--- a/src/layout/Header.js
+++ b/src/layout/Header.js
@@ -23,7 +23,7 @@ class Header extends React.Component {
 
         var isRowOrTab = false;
         React.Children.forEach(this.props.children, child => {
-            if(child.type === HeaderRow || child.type === HeaderTabs) {
+            if(child && (child.type === HeaderRow || child.type === HeaderTabs)) {
                 isRowOrTab = true;
             }
         });


tleunen:
Not sure about this one... If the component doesn't have any children, it should not go inside the forEach loop.

Graf009:
Example:
<Header ...>
    {!userIsLogin && <Button .....>LOG IN</Button>}    // return false
</Header>

tleunen:
Hmm. Would need to test this one myself. Could be a React issue then, I'm not sure we should be able to loop over a `false` child. (https://facebook.github.io/react/tips/false-in-jsx.html)

Graf009:
```javascript
<Header transparent title="Example">
        {!userIsLogin && <Button raised colored onClick={ onLoginButton }>LOG IN</Button>}
</Header>
```
In article https://facebook.github.io/react/docs/component-specs.html says "You can also return null or false to indicate that you don't want anything rendered."
I think it's safe behavior, what do you say?

tleunen:
I had a crazy weekend so I didn't get the chance to take a look at this.
Would probably need to put a breackpoint in this `traverseAllChildren` and see what's going on :)
https://github.com/facebook/react/blob/master/src/shared/utils/traverseAllChildren.js

tleunen:
I finally got time to check this one, and I can not reproduce the issue. If the Header doesn't have any children, the children property is undefined and therefore it goes never inside the children.foreach function.

Graf009:
Here is an example that reproduces the error https://gist.github.com/Graf009/ddef4109d157046ac62a

PR #23, opened 2015-08-28
Fix Grid.js bug for phone sizes

slybridges:
Copy/Paste bug I guess :-)


tleunen:
Good catch! ;)
I now understand why it didn't seem to work when I used the `phone` prop...

slybridges:
Yep that's what I scratched my head around too for some time ;-)

PR #24, opened 2015-08-31
Make possible to install from github

web2style:
I agree that the src folder should be included with the module. If you agree, you can consider the following pull request. Now it compiles locally after installation.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,2 @@
-src/
 demo/
 webpack.config.js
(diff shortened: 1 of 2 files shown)


tleunen:
`postinstall` will fail because babel is set as `devDependencies`. And I don't want to put it as a direct dependency of the project.

But I agree we can put src/ on npm as well.

I think that's all we can do. Usually people won't install the project through the github link but with the npm name, so they won't have the issue. For people that get the code from github, they can either fork it and handle it the way they want, or they can also put a postinstall hook in their project to install the dependencies of react-mdl as well.

Regarding this project, I had a crazy weekend so I could not release a new version but will do asap

PR #25, opened 2015-09-02
Get more data from onChange event

web2style:
Now you will be able to do smth like this:

```
<Switch
    id="shortcuts"
    checked={this.state.shortcuts}
    onChange={this.handleChange}
/>

handleChange(e) {
    this.setState({
      [e.target.id]: e.target.checked
    })
  }
```

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -18,7 +18,6 @@ class Switch extends React.Component {
 
     render() {
         var { checked, className, disabled, id, ripple } = this.props;
-        var inputId = 'switch-' + id;
 
         // enable ripple by default
         ripple = ripple !== false;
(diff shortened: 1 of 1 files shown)


tleunen:
What are the others properties of the event? Are they useful or could we just send `e.target`?

web2style:
There are a lot of other properties. At least 2 of them are useful for me: isDefaultPrevented: () and
isPropagationStopped: (). I think there's no sense to pass e.target and limit users in these properties. At least if you don't want to accept another pull request in the future :)

web2style:
I think this pull request was a hasty decision. First of all, this is a braking change. Also, other components send boolean as onChange payload, so sending an object here will be an inconsistency. So I think you should discard this commit.

What concerns my example above, it can be handled like this:

```
(checked) => this.handleChange('yourId', checked)
```

tleunen:
The value for `onChange` really depends on the component.. Some will return a single value (string for Textfield, boolean for checkbox) so this one could send whatever he wants... The object or the combo "id, checked" could be another solution.

I don't mind changing it again for "id, checked" if you want, but in that case, I would change it as "checked, id" so it's not a breaking change (But again, creating a BC now is not a problem since we're still not in 1.0)

web2style:
I didn't mean you should pass id as an argument from inside the component. A user can pass it as a string himself in the onChange callback when he creates a components if he needs it.

I just proposed to discard this commit and get back to how it was before. But it is up to you anyway.

PR #35, opened 2015-09-16
Tab ripple

### LINKED ISSUE #27: Tabs should have option to ripple
Pleasure to use this library. Just wondering if tab rippling is a planned feature? I tried just adding `mdl-js-ripple-effect` to the tabs, evidently there is more to it than that. 

react-mdl Tabs demo:
noripple

[getmdl.io Tabs demo](http://www.getmdl.io/components/index.html#layout-section/tabs):
ripple

timkrins:
To fix #27

Removed unrequired function calls in material.js

Added demo

### DIFF
--- a/extra/material.js
+++ b/extra/material.js
@@ -2457,12 +2457,6 @@ function MaterialTab(tab, ctx) {
         }
         tab.addEventListener('click', function (e) {
             e.preventDefault();
-            var href = tab.href.split('#')[1];
-            var panel = ctx.element_.querySelector('#' + href);
-            ctx.resetTabState_();
-            ctx.resetPanelState_();
-            tab.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
-            panel.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
         });
     }
 }
--- a/extra/material.js
+++ b/extra/material.js
@@ -2457,6 +2457,12 @@ function MaterialTab(tab, ctx) {
         }
         tab.addEventListener('click', function (e) {
             e.preventDefault();
+            var href = tab.href.split('#')[1];
+            var panel = ctx.element_.querySelector('#' + href);
+            ctx.resetTabState_();
+            ctx.resetPanelState_();
+            tab.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
+            panel.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
         });
     }
 }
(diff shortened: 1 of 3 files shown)


extra/material.js
```diff
@@ -2457,12 +2457,6 @@ function MaterialTab(tab, ctx) {
         }
         tab.addEventListener('click', function (e) {
             e.preventDefault();
-            var href = tab.href.split('#')[1];
-            var panel = ctx.element_.querySelector('#' + href);
-            ctx.resetTabState_();
-            ctx.resetPanelState_();
-            tab.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
-            panel.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
```
tleunen: The problem here is that we try to avoid making modifications to the base MDL library. And the code you're removing here is the reason why the Tab(s) component doesn't implement the `upgrade` function.

It would be better to find another solution to resolve the issue without making any modifications to the parent library. If this is not possible, it would be interesting to ask on the MDL repository if we can find a better way to handle this case. Maybe just add a check on the href before trying to reset the panels should be enough and maybe they will accept this small change.

timkrins: Yeah, I was trying to avoid making changes, but noticed that the current diff between the two material.js files removed similar functions that are obsolete in React style development.

As it is, the changes I made will work with the current material.js file, but causes errors in the console if I don't have a tab panel and hash-links configured.

I will do some thinking on the problem :)

tleunen: tbh, I think the best thing to do here is submit a PR in mdl to add a check before reseting everything. (it would be a safe check for them but 
And then we should be able to upgrade the element :)

timkrins: Hmm, would mdl want this changed in their branch? It will currently fail if there is no dom panel for mdl to make visible, and if you are using pure mdl, this is probably an error that you want to see and then correct.
Like this person's issue which you have commented on before: google/material-design-lite#1397
Maybe the upstream pull request could implement a warning that there was no panel found? But then I'm not sure how we would ignore that warning...

tleunen:
Nice hack. This is not optimal because we would get multiple container with the same id "`undefined`" if we put more than 1 tabs component in the page, but I guess we can live with that for now...

We should not put any anchor on the tabs so we should be good on that side.
I'll keep open for now and will test the code asap before merging.

tleunen:
Thanks a lot @timkrins, I squashed the commits and manually merged it.

PR #38, opened 2015-09-18
Add 'invalid' props to Textfield for external error checking

### LINKED ISSUE #37: Textfield: add props for setting externally checked validity of field
For now, field errors can only be triggered by the `pattern` props. But `pattern` checks are limited (for instance. we cannot ask a server for field validity).

I found it useful to add an `invalid` props to Textfield that triggers displaying `error` props when it is set to `true`

Would you be willing to accept such PR ?

slybridges:
Closes #37


slybridges:
Closing this one and will open a new one after our discussion.

Issue #40, opened 2015-09-21
Custom component for buttons


faergeek:
It would be great to allow passing needed component as a prop to use something other than `button`.

For example I'd like to have a link that looks like a button and would be cool if `react-mdl` could handle that.
See https://facebook.github.io/react/docs/animation.html#rendering-a-different-component

Also would be cool if it could "guess" what I'm trying to do based on passed props, i.e. if I pass `href` then I want it to render `a` tag.
See http://react-bootstrap.github.io/components.html#buttons-tags

Some examples:
```javascript
<Button>button</Button>
<Button component="a">anchor tag</Button>
<Button href="#">anchor tag</Button>
<Button component={Link}>Link component from react-router</Button>
```

If you like the idea, but you're busy I could submit pull request.

tleunen:
I really like the idea. And yes you can submit a PR for that. I'm still quite busy these days :/
Thanks!

faergeek:
I think actually it may be needed in many other places. May be we should write some mixin-link solution for that.

PR #43, opened 2015-09-21
Allow to override element that Button creates

faergeek:
Fix for #40.

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -7,6 +7,8 @@ class Button extends React.Component {
         accent: PropTypes.bool,
         className: PropTypes.string,
         colored: PropTypes.bool,
+        component: PropTypes.any,
+        href: PropTypes.string,
         primary: PropTypes.bool,
         raised: PropTypes.bool,
         ripple: PropTypes.bool
@@ -18,7 +20,8 @@ class Button extends React.Component {
 
     render() {
         var { accent, className, colored,
-            primary, raised, ripple, ...otherProps } = this.props;
+            primary, raised, ripple, component, href,
+            children, ...otherProps } = this.props;
 
         var buttonClasses = classNames('mdl-button mdl-js-button', {
             'mdl-js-ripple-effect': ripple,
@@ -28,11 +31,13 @@ class Button extends React.Component {
             'mdl-button--accent': accent
         }, className);
 
-        return (
-            <button className={buttonClasses} {...otherProps}>
-                {this.props.children}
-            </button>
-        );
+        component = component || (href ? 'a' : 'button');
+
+        return React.createElement(component, {
+            className: buttonClasses,
+            href,
+            ...otherProps
+        }, children);
     }
 }
 


faergeek:
What about that PR?

src/Button.js
```diff
@@ -7,6 +7,8 @@ class Button extends React.Component {
         accent: PropTypes.bool,
         className: PropTypes.string,
         colored: PropTypes.bool,
+        component: PropTypes.any,
```
tleunen: Shouldn't this be `PropTypes.element` ? Or if we want to allow string as well (but not sure we want that), we could do a `oneOfType`...

src/Button.js
```diff
@@ -7,6 +7,8 @@ class Button extends React.Component {
...
+        href: PropTypes.any,
```
tleunen: Use `PropTypes.string` instead

src/Button.js
```diff
@@ -14,7 +16,8 @@ class Button extends React.Component {
 
     render() {
         var { accent, className, colored,
-            primary, raised, ripple, ...otherProps } = this.props;
+            primary, raised, ripple, component, href,
```
tleunen: Actually, no need to get the `href` since it will be inside `otherProps`. Since we do not make any changes to the value, that's fine.

But this could give an `eslint` error since it will be defined as a prop but "never" used? Could you test?

src/Button.js
faergeek: Nope, `PropTypes.element` would be already rendered element, but we should allow to pass a string (for html tags), class (for react components) and soon function (for react 0.14 feature to allow just pass function).

So we actually need to use `oneOfType` with `string` and `func`, but I think we should not restrict it. React 0.14 will allow to use pure function instead of a real class and there's so much issues with different components out there that did not assume that possibility and it's just annoying.

Anyway react will warn users if they'll try to pass something totally wrong.

src/Button.js
faergeek: I think we should not restrict here too. What if there's some component which use a prop called `href` and it can be function or node or something.

For example, what if I'll write some `ScrollToElement` component which will scroll to some element. It can pass function to return element id or something like this.

It's strange case though. What do you think?

src/Button.js
faergeek: We get `href` just to check that it is set later. So we can set `a` tag for button if user didn't pass custom component.

P.S. for now `eslint` doesn't warn about unused props that are in props validation. But I wish it would :-)

src/Button.js
faergeek: BTW, take a look at `TransitionGroup` component which is included in react in 0.13.

tleunen: :+1:

src/Button.js
tleunen: Yep but I think it's not valid for `href`. `href` has always been used as a string, so people know they should put a string in there, nothing else. If you write a custom element taking a non string `href`, maybe you should rename it ;)

src/Button.js
tleunen: You're right ;)

faergeek:
I fixed `href` validation. If everything alright, I can just squash it.

tleunen:
Yep, a squash would be nice :)

faergeek:
Done :-)

PR #45, opened 2015-09-21
Use defaultProps instead of falling back in render

faergeek:
I think it's more clear to use `defaultProps` instead of falling back in render, because you understand default values better looking at the top of component code.


tleunen:
Thanks ;)

Issue #47, opened 2015-09-21
Some components may not upgrade on remount


faergeek:
I noticed that some components (in my case tooltips) do not work after remount.

After some investigation I noticed that mdl expects root node of component as an argument, but tooltip returns wrapper around that node.

So I think we should try to upgrade/downgrade DOM node found using something like this:

```javascript
React.findDOMNode(this.refs.mdlComponent || this)
```

Graf009:
I think that it is necessary for all the components through the use of links refs

faergeek:
It's really needed just for Tooltip, for others `React.findDOMNode(this)` is enough for now.

Graf009:
I use the react-modal, and I have problems with the same problems React.findDOMNode (this).

tleunen:
fixed by #48

PR #48, opened 2015-09-21
Better way to upgrade/downgrade mdl components

faergeek:
Here I'm adding ref `mdlComponent` to components which render wrappers around mdl components (actually it's just Tooltip).

Another change here is wrapping methods instead of wrapping whole components. Aside from easier access to `this.refs` it now looks better in `react-devtools`.

### DIFF
--- a/src/utils/mdlUpgrade.js
+++ b/src/utils/mdlUpgrade.js
@@ -1,19 +1,12 @@
 import React from 'react';
+import MDLComponent from './MDLComponent';
 
-export default (Component) => {
-    class MDLUpgradedComponent extends React.Component {
-        componentDidMount() {
-            componentHandler.upgradeElement(React.findDOMNode(this));
-        }
+export default Component => {
+    const render = Component.prototype.render;
 
-        componentWillUnmount() {
-            componentHandler.downgradeElements(React.findDOMNode(this));
-        }
+    Component.prototype.render = function() {
+        return <MDLComponent>{this::render()}</MDLComponent>;
+    };
 
-        render() {
-            return <Component {...this.props} />;
-        }
-    }
-
-    return MDLUpgradedComponent;
+    return Component;
 };
--- a/src/ProgressBar.js
+++ b/src/ProgressBar.js
@@ -1,5 +1,6 @@
 import React, { PropTypes } from 'react';
 import classNames from 'classnames';
+import mdlUpgrade from './utils/mdlUpgrade';
 
 class ProgressBar extends React.Component {
     static propTypes = {
@@ -10,8 +11,6 @@ class ProgressBar extends React.Component {
     }
 
     componentDidMount(){
-        // Can't use the mdlUpgrade component because we need access to the Material internal object
-        componentHandler.upgradeElement(React.findDOMNode(this));
         this.setProgress(this.props.progress);
         this.setBuffer(this.props.buffer);
     }
@@ -19,10 +18,6 @@ class ProgressBar extends React.Component {
     componentDidUpdate() {
         this.setProgress(this.props.progress);
         this.setBuffer(this.props.buffer);
-    }
-
-    componentWillUnmount(){
-        componentHandler.downgradeElements(React.findDOMNode(this));
     }
 
     setProgress(progress) {
@@ -48,4 +43,4 @@ class ProgressBar extends React.Component {
     }
 }
 
-export default ProgressBar;
+export default mdlUpgrade(ProgressBar);
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -1,8 +1,8 @@
 import React, { PropTypes } from 'react';
 import classNames from 'classnames';
-import mdlUpgrade from './utils/mdlUpgrade';
+import MDLComponent from './utils/MDLComponent';
 
-class Tooltip extends React.Component {
+export default class Tooltip extends React.Component {
     static propTypes = {
         children: PropTypes.oneOfType([
             PropTypes.element,
@@ -33,16 +33,16 @@ class Tooltip extends React.Component {
 
         return (
             <div style={{display: 'inline-block'}} {...otherProps}>
-                {React.cloneElement(element, {
-                    id: id
-                })}
-                {React.cloneElement(label, {
-                    htmlFor: id,
-                    className: classNames('mdl-tooltip', { 'mdl-tooltip--large': large })
-                })}
+                {React.cloneElement(element, { id })}
+                <MDLComponent>
+                    {React.cloneElement(label, {
+                        htmlFor: id,
+                        className: classNames('mdl-tooltip', {
+                            'mdl-tooltip--large': large
+                        })
+                    })}
+                </MDLComponent>
             </div>
         );
     }
 }
-
-export default mdlUpgrade(Tooltip);
--- a/src/utils/MDLComponent.js
+++ b/src/utils/MDLComponent.js
@@ -1,0 +1,15 @@
+import { Children, Component, findDOMNode } from 'react';
+
+export default class MDLComponent extends Component {
+    componentDidMount() {
+        componentHandler.upgradeElement(findDOMNode(this));
+    }
+
+    componentWillUnmount() {
+        componentHandler.downgradeElements(findDOMNode(this));
+    }
+
+    render() {
+        return Children.only(this.props.children);
+    }
+}


faergeek:
It fixes #47

tleunen:
You are now modifying the component instead of creating a new one. I'm not sure we want to go in that direction...
I would prefer if we can find another solution to the issue.

faergeek:
Okay, so what about inheritance? Something like this:
```javascript
class ProgressBar extends MDLComponent {
...
}
```
or another way around (new component inherits from ours, less changes in code, but looks not so good in devtools :-( ):
```javascript
export default Component =>
class MDLComponent extends Component {
...
}
```

tleunen:
Not a big fan of inheritance when writing React code either.

faergeek:
Are there some reasons for that? It fits naturally in that case.

tleunen:
Inheritance in JavaScript is hard because there's no way to know when you override a function or not. That's why I usually try to not use it (except in some cases).
Before going with that route, I'll try to find another way of solving this.

faergeek:
What do you mean by "override a function or not"?

tleunen:
Inside a child class, you have no way to know without looking at the parent if a function you're writing is overriding the parent function. That's why I tend to avoid using inheritance in JS.

faergeek:
We can use `super` to call parent
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super

faergeek:
Let me provide a prototype to play with it.

faergeek:
It's time to check it out.
Sorry for delay.

faergeek:
Ping

tleunen:
Still thinking of a solution. I'd like to keep the High order component instead of the inheritance so I'm trying to find a way to fix the issue while keeping the HoC.

faergeek:
What if we will use `MDLComponent` **inside** components' `render` methods like this?

```javascript
class Button extends React.Component {
    render() {
        var { accent, className, colored,
            primary, raised, ripple, ...otherProps } = this.props;

        var buttonClasses = classNames('mdl-button mdl-js-button', {
            'mdl-js-ripple-effect': ripple,
            'mdl-button--raised': raised,
            'mdl-button--colored': colored,
            'mdl-button--primary': primary,
            'mdl-button--accent': accent
        }, className);

        return (
            <MDLComponent>
              <button className={buttonClasses} {...otherProps}>
                  {this.props.children}
              </button>
            </MDLComponent>
        );
    }
}
```

`MDLComponent` should wrap exactly one children on which we should call `upgradeElement`.
It will provide good public API for users wishing to change real markup however they want.

For example:

```javascript
        return (
            <MDLComponent>
              <input
                type="button"
                className="mdl-button mdl-js-button"
                value="Send the from"
              >
            </MDLComponent>
        );
```

tleunen:
What do you think of adding a get function inside the components who needs to manually specify the container to downgrade, and update the `mdlUpgrade` helper with this:
```
componentWillUnmount() {
    var { comp } = this.refs;
    var element = (comp.getMdlContainer && comp.getMdlContainer()) || this;
    componentHandler.downgradeElements(React.findDOMNode(element));
}

render() {
    return <Component ref="comp" {...this.props} />;
}
```

faergeek:
I think that we should not use refs for that. I really like the idea with HoC which wraps mdl components.
For example `Tooltip`'s `render` would look something like that:

```javascript
    render() {
        var { label, large, children, ...otherProps } = this.props;
        var id = Math.random().toString(36).substr(2);

        if(typeof label === 'string') {
            label = <span>{label}</span>;
        }

        var element;
        if(typeof children === 'string') {
            element = <span>{children}</span>;
        }
        else {
            element = React.Children.only(this.props.children);
        }

        return (
            <div style={{display: 'inline-block'}} {...otherProps}>
                {React.cloneElement(element, { id })}
                <MDLUpgrade>
                    {React.cloneElement(label, {
                        htmlFor: id,
                        className: classNames('mdl-tooltip', {
                            'mdl-tooltip--large': large
                        })
                    })}
                </MDLUpgrade>
            </div>
        );
    }
```

Even better for devtools we can do that without even leaving the traces of HoC like in `react-document-title`
https://github.com/gaearon/react-document-title/blob/75ef9566a162a4c9bc697c0e23cf5c44ac5efb74/index.js#L22-L28

faergeek:
Wrong, it leaves the traces in devtools, but anyway I think it's a good solution.

tleunen:
Hmm.. Let's keep the old way of upgrading (the function mdlUpgrade), and add a new `MDLUpgrade` for this specific tooltip code.

faergeek:
If you do not want so much changes I can rewrite old `mdlUpgrade` which will use HoC and we can leave the code of most components as it was before.

Do you see any other problem?

tleunen:
I was more thinking of keeping the `mdlUpgrade` as it was before, and creating a new `MDLUpgrade` component that will use the `mdlUpgrade` function. That way we keep the logic only in one location.

faergeek:
I just thought the other way around.
In that case we can avoid workarounds like in ProgressBar component

faergeek:
What you think about exposing `mdlUpgrade` as a public function?
Also what you think about it's usage as a decorator instead of plain function call?

tleunen:
One or the other, it will give the same result. It's just that one needs less code changes :)
About the decorator, I prefer keeping a function for now, less code, easier to understand for everyone.

faergeek:
It's less code indeed.

```javascript
class ProgressBar extends React.Component {
}

export default mdlUpgrade(ProgressBar);
```

vs

```javascript
@mdlUpgrade
export default ProgressBar extends React.Component {
}
```

and no other changes needed.

tleunen:
I meant, less code change to keep the function as is and create a component.

Using a generator will produce more code in the end. We might switch to a generator later, but not now

faergeek:
Please note that changes in `ProgressBar` and explicit use of `MDLUpgrade` component is needed as a workaround instead of what was before.

It's just some small code share.

faergeek:
And if everything's alright I'll squash it again.

index.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
     Spacer: Layout.Spacer,
     Navigation: Layout.Navigation,
     Content: Layout.Content,
+    MDLUpgrade: require('./lib/MDLUpgrade'),
```
tleunen: Why would we export the component?

src/MDLUpgrade.js
```diff
@@ -0,0 +1,15 @@
+import { Children, Component, findDOMNode } from 'react';
+
+export default class MDLUpgrade extends Component {
```
tleunen: Seems like it should go inside the `utils` directory

index.js
faergeek: I thought we could export it for users wishing some custom markup or something like this.
For example I use code like this if I want link look like a button:

```javascript
                  <a
                    className="mdl-button mdl-button--icon"
                    href={`/download/${id}`}
                    title="Download"
                  >
                    <i className="material-icons">file_download</i>
                  </a>
```

Would be handy if in such cases user could wrap that code inside `MDLUpgrade` which will handle remounting.

tleunen: But for your example, you could use the component we provide as well.

Not sure the export is needed. Because if the end-user needs access to it, it means something is missing in the library ;)

src/MDLUpgrade.js
faergeek: Sorry, I thought that if it's a component and not function which wraps component then I should put it alongside other components.

index.js
faergeek: Ok, I'll fix it.

src/MDLUpgrade.js
tleunen: Yep, but it's more of a "helper" than a real component :)

faergeek:
I had to rename `MDLUpgrade` to `MDLComponent` to avoid collisions with `mdlUpgrade` on case-insensitive filesystems.

tleunen:
make sense. :+1:

src/utils/mdlUpgrade.js
```diff
@@ -1,19 +1,9 @@
...
-            componentHandler.downgradeElements(React.findDOMNode(this));
-        }
-
-        render() {
-            return <Component {...this.props} />;
-        }
+export default Component =>
+class MDLUpgradedComponent extends React.Component {
+    render() {
+        return <MDLComponent><Component {...this.props} /></MDLComponent>;
```
tleunen: So for components using this helper, there are now 2 indirections:
```
MDLUpgradedComponent
    MDLComponent
        Component
```
While if we reverse the helper function and component, we'll have only 1 indirections in most cases.

faergeek: Sadly, yes. I'll try it out a little later.

faergeek:
Finally I coded something. The reason why I don't want to reverse `mdlUpgrade` and `MDLComponent` is I don't want to lose `this` and use workarounds as was in `ProgressBar`.

tleunen:
I'll merge.. we'll see how it goes like that. Even if I'm not a big fan of the `mdlUpgrade` implementation, I currently don't have any better solution

faergeek:
Thanks

Issue #42, opened 2015-09-21
Textfields missing 'is-invalid' when error is set.


Pagebakers:
Errors are never shown because the `is-invalid` className is not added when an error is set.

tleunen:
For now, the `is-invalid` state is only set when the value doesn't match the provided `pattern`. There's some discussions about the textfield here https://github.com/tleunen/react-mdl/issues/37
We're trying to find the best available solution to resolve this.

tleunen:
Fixed by merging #49

kctang:
These fields:

```javascript
<Textfield name="check1" label="Email" error="something wrong" pattern="" defaultValue="bla"/>
<Textfield name="check2" label="Email" error="this error message not showing" defaultValue="bla"/>
```

Current behavior on component render:
* check1 shows error.
* check2 does not show error

Would be better if `check2` also show error because pattern is not set.

tleunen:
I thought we were handling the case but I guess not. Could you please open a new ticket to track the issue?

jolic:
Wouldn't it be good to add a callback when an input item isn't valid?

Something like:
```
<Textfield
                onChange = {() => {}}
                pattern='....'
                error="Show this error"
                label="Test"
                isValid = {(isValid) => { console.log("This item is: "+valid); } }
            />
```

Currently only the parent element of the input gets the classname "is-invalid".

tleunen:
I think you can use the `onInvalid` event, or the `onChange` event and check the validity of the input field there.

PR #49, opened 2015-09-22
Force `error`props display if `pattern` isn't set

### LINKED ISSUE #37: Textfield: add props for setting externally checked validity of field

slybridges:
Fixes #37 and #42

### DIFF
--- a/demo/textfield/textfield.js
+++ b/demo/textfield/textfield.js
@@ -18,8 +18,14 @@ class Demo extends React.Component {
             textfield3: '',
             textfield4: '',
             textfield5: '',
-            textfield6: ''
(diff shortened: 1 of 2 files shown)


tleunen:
The solution looks nice, at least until the MDL team provide a better way of doing this.

Thanks.

Pagebakers:
Great, thanks!

PR #56, opened 2015-09-28
Do not provide onChange prop by default

### LINKED ISSUE #55: LinkState and valueLink bug
When I use the linkState mixin:

Uncaught Error: Invariant Violation: Cannot provide a valueLink and a value or onChange event. If you want to use value or onChange, you probably don't want to use valueLink.

Any idea?

faergeek:
This PR fixes #55

### DIFF
--- a/src/Textfield.js
+++ b/src/Textfield.js
@@ -24,10 +24,6 @@ class Textfield extends React.Component {
         ])
     }
 
-    static defaultProps = {
-        onChange: function() {}
-    }
-
     componentDidUpdate(prevProps) {
         if(
             this.props.required !== prevProps.required ||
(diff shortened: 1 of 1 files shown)


src/Textfield.js
```diff
@@ -60,15 +52,19 @@ class Textfield extends React.Component {
...
             key: inputId,
             value: value,
-            onChange: this._handleChange,
             rows: rows,
             ...otherProps
-        });
+        };
+
+        if (onChange) inputProps.onChange =
+            event => this.props.onChange(event.target.value);
```
tleunen: Creating an anomymous function every render is not optimal. It's way better to keep the `_handleChange` ;)

faergeek: Sorry, I just got used to use `react-hot-loader` and now `react-transform-hmr`. So I avoid assigning methods on each instance.

tleunen: Let me know when it's fixed and I'll merge

faergeek:
Fixed

PR #58, opened 2015-09-29
Support for programmatic change of Textfield value

faergeek:
This will correctly work with floatingLabel when
change from empty value or to empty value

### DIFF
--- a/src/Textfield.js
+++ b/src/Textfield.js
@@ -28,7 +28,8 @@ class Textfield extends React.Component {
     componentDidUpdate(prevProps) {
         if(
             this.props.required !== prevProps.required ||
-            this.props.pattern !== prevProps.pattern
+            this.props.pattern !== prevProps.pattern ||
+            this.props.value !== prevProps.value
         ) {
             findDOMNode(this).MaterialTextfield.checkValidity();
         }
@@ -40,6 +41,9 @@ class Textfield extends React.Component {
             // If we want to force the error display, we have to override mdl 'is-invalid' value.
             var elt = findDOMNode(this);
             elt.className = classNames(elt.className, 'is-invalid');
+        }
+        if (this.props.value !== prevProps.value) {
+            findDOMNode(this).MaterialTextfield.checkDirty();
         }
     }
 


tleunen:
What's the usecase?
Would have thought that because the value has changed, the textfield is re-rendered and therefore MDL should check the validity.

faergeek:
Try to update value using react-devtools from empty to some value and backwards. If value assigned programmatically, then label will be under real text and vice versa.

faergeek:
I have a real use case where I'm editing images and when I switch them some images have filled values and some don't.

faergeek:
Not sure if it's needed for validation, though.

tleunen:
Good catch.
Actually, this is because the MDL object is already initialized.
We might want to use `.change()` on the MDL object if we detect the value has changed.

faergeek:
I tried using `.change()`, seems like it doesn't work that way. Did you test it?

tleunen:
Not yet no... But it should work since it updates the css.

faergeek:
Usually such methods assume that you know what you're doing and handled everything, so it does not dispatch event.

faergeek:
Ping

faergeek:
@tleunen So what about that PR?

tleunen:
Did not get a chance to test it yet. Should do that tomorrow.

faergeek:
Just did a rebase on master.
Also I don't see any way to trigger `change` event at all.

tleunen:
So I checked it, with react-devtools, and there one thing I don't understand... When changing the value, the `textfield` component receive an event in `componentDidUpdate` where the `prevProps` are the props with the value I entered, and `props` empty.

But it should be the opposite.

faergeek:
If you're testing this in demos then you can just see other fields updating too as there are rerender of parent component and we have no `shouldComponentUpdate` (and we really should) for them.

I checked it, everything is alright, except we should add `shouldComponentUpdate` hook here and I think in more places across library (will make another PR for that).

faergeek:
I just did a rebase on top of master.

faergeek:
Can we talk about that problem in a more real time way? May be gitter or something like that?

tleunen:
I created the gitter room so we can discuss about it :)

tleunen:
closing this in favor of #79

PR #59, opened 2015-09-29
Change peerDependencies to support beta and newer

faergeek:
Without that PR there will be warning when you install `react@0.14.0-beta3` or `react@0.14.0-rc1`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "webpack-dev-server": "1.12.0"
   },
   "peerDependencies": {
-    "react": "0.13.x || ^0.14.0"
+    "react": "0.13.x || ^0.14.0-alpha1"
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 0",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   "devDependencies": {
     "babel": "5.8.23",
     "babel-loader": "5.3.2",
+    "react": "0.13.3",
     "webpack": "1.12.2",
     "webpack-dev-server": "1.12.0"
   },


tleunen:
Hmm.. What do we have to set to accept all 0.14 versions?

faergeek:
I guess first [publicly announced](https://facebook.github.io/react/blog/2015/07/03/react-v0.14-beta-1.html) beta was `beta1`. I think that would be enough.

We can go from `0.14.0-alpha1` to support all releases, but I think that now even just `rc1` would be enough. AFAIK there's no breaking changes between `beta1` and `rc1`.

So I think we should set `beta1` or `rc1`.

tleunen:
`>= 0.13 < 0.15`, will that solve the issue?

faergeek:
I guess. Need to try it out with both versions and check if npm shows warnings about unmeet peer dependency.

tleunen:
We also want to set React as a devDepencency to prevent issues with npm3 :)
Will make a few tests later tonight

faergeek:
No, having react as a dependency is the first step to include two versions of react. And it's bad and not working at all.
What issues do you have with `npm@3`? I don't have any.

faergeek:
If you mean we should have it for tests, then I think we should find some another way for that.

tleunen:
That's why I said "**dev**Dependency" :)

faergeek:
Hmm, looks like redux does include `react` in `devDependencies` too.
https://github.com/rackt/react-redux/blob/master/package.json#L53

tleunen:
The reason for this is simple. NPM doesn't install peerDependency anymore, so you have to also set it as a devDependency so you can still get the project, run the tests/demos ;-)

faergeek:
Yeah, that makes sense :+1: 
Do you have any idea on how we can get rid of deprecation warnings in `react@0.14.x` about `findDOMNode`?

tleunen:
Removing them once 0.14 is released :) In the meantime, I'm afraid there is no solution for that.

I was also thinking of creating a `next` branch of react-mdl based on the next version of react

faergeek:
I think creating another branch doesn't make sense because it doesn't break anything actually.

We can add some info about `0.14.x` support to readme though.

faergeek:
Updated

PR #75, opened 2015-10-13
Create mdl base class in place of util methods

euforic:
- Create MDLUpgradedComponent base class
- Create MDLComponent base class
- Update upgraded components to use new MDLUpgradedComponent base class
- Expose MDLComponent and MDLUpgradedComponent classes
- Remove no longer needed util files

This pr will also solve the problem of upgraded components adding an extra wrapper not giving direct access to the components root element


faergeek:
Please read #48

tleunen:
Yes as @faergeek mentioned, I'm not in favor or inheritance in this case. That's why we decided to use another approach :)

PR #78, opened 2015-10-15
Doesn't render the badge if there's no children

### LINKED ISSUE #76: Badge empty text case
If in result of component render we've got empty attribute in DOM now, attribute should not be rendered at all.

So it should not be required too.

tleunen:
fixes #76

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -5,12 +5,16 @@ class Badge extends React.Component {
         children: PropTypes.oneOfType([
             PropTypes.element,
             PropTypes.string
-        ]).isRequired,
+        ]),
         text: PropTypes.string.isRequired
     }
 
     render() {
         var { children } = this.props;
+
+        // No badge if no children
+        if(!React.Children.count(children)) return null;
+
         var element;
         if(typeof children === 'string') {
             element = <span>{children}</span>;


faergeek:
Sorry, it's not what I'm talking about. It's about `text` prop, not `children`.

The point is that we're wrapping some element in `Badge` and we can provide `text`, but according to upstream styles the element is shown only if DOM element has a `data-badge` attribute, so if it's empty (there's no new notifications or something like that) then it doesn't make sense to render it.

faergeek:
Here's a selector `.mdl-badge[data-badge]:after`.

tleunen:
Haha ok ;) Yep, actually I figured out it make no sense if the children was null as well anyway ;)
About the text, I'm not sure.. How would you render an empty badge? (just the red circle without any content)?
I'm not saying there's no fix to be made on react-mdl. But maybe we could render an empty data-badge if text is null? Because if you don't want the badge, you could not render it in your component instead of assuming Badge will not render himself

faergeek:
Let's assume that we have some icon for notifications and we want to show a badge only if there's a notification (mostly like on github, but with numbers), with the current implementation we will wrap component in a badge if there's notifications or don't wrap if there's none. Not so good.

Instead I guess it's assumed by upstream that if we don't want to show badge we just don't set `data-badge` attribute, even empty.

I can make a PR to show the difference.

tleunen:
Ok so you suggest that if `text` is null, `Badge:render` returns the children then?

faergeek:
Not exactly. I'll make a PR to explain better :-)

faergeek:
It throws now when you return `null` from `Badge`.

faergeek:
What if instead of checking for a single element and so on and so forth we just wrap element in span in any case? Try to wrap `Textfield` in `Badge`, for example. It's confusing.

tleunen:
No need to add an extra container for that.
What do you mean when you return `null` from `Badge`?

src/Badge.js
```diff
@@ -5,12 +5,16 @@ class Badge extends React.Component {
...
-        ]).isRequired,
+        ]),
         text: PropTypes.string.isRequired
     }
 
     render() {
         var { children } = this.props;
+
+        // No badge if no children
+        if(!React.Children.count(children)) return null;
```
faergeek: I'm talking about this line.
It seems like stateless components do not support returning null from render.
Current test suite skips it somehow.

tleunen: Hmm weird. Do you know if there's already a bug about that on react?
Seems weird the behavior from a stateless component is different in React and in a Shallow renderer.

faergeek: I guess may be we just don't get stateless components right?)
I didn't search for a bug.

PR #79, opened 2015-10-15
fixes textfield not being updated if it received dynamically a value

tleunen:
Related to #58.

If the textfield receives dynamically a value, without directly through the input from a user, the textfield wasn't updated properly (the floating label/css classes were not updated).
By forcing to change the value with the MDL upstream `change` function, it resolves the issue.

### DIFF
--- a/src/Textfield.js
+++ b/src/Textfield.js
@@ -35,6 +35,9 @@ class Textfield extends React.Component {
         if(this.props.disabled !== prevProps.disabled) {
             findDOMNode(this).MaterialTextfield.checkDisabled();
         }
+        if(this.props.value !== prevProps.value && this.refs.input !== document.activeElement) {
+            findDOMNode(this).MaterialTextfield.change(this.props.value);
+        }
         if(this.props.error && !this.props.pattern) {
             // At every re-render, mdl will set 'is-invalid' class according to the 'pattern' props validity
             // If we want to force the error display, we have to override mdl 'is-invalid' value.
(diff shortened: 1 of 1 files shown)


faergeek:
Still to avoid setting value twice we should not pass argument to `change` method.

tleunen:
So I updated the code to only force a `change` value on the textfields that are not focused.
What do you think?

faergeek:
I don't think we need these checks actually :-) Just leaving it with no argument will be ok.

Take a look at these lines https://github.com/tleunen/react-mdl/blob/master/extra/material.js#L2637-L2639

So if we don't pass a value it just updates classes.
And as `classList` is used everywhere we won't have any actual changes when they're not needed and no performance problems.
Take a look at this video for explanation why http://www.youtube.com/watch?v=hZJacl2VkKo

faergeek:
Even better, upstream code checks if there's need to add/remove classes, so even `classList` is not touched.

tleunen:
Yes, it would work, but only with the current version that we have in react-mdl. Upstream, the code has changed to fix an issue and now it supports to pass an empty string.
https://github.com/google/material-design-lite/blob/master/src/textfield/textfield.js#L201-L205

faergeek:
Even with that in mind I don't that we should use contains to detect if it's focused, it's better to use ref and just check equality then :-)

But don't close this yet, I'd like to test if setting value twice will have any impact on performance.

tleunen:
:+1: used ref instead to avoid searching inside the entire element.

faergeek:
Could you publish that to npm before we introduce breaking changes?

tleunen:
There are already too many changes and breaking changes, so I was planning to finish them before releasing a new version

PR #81, opened 2015-10-16
Allow empty text on Badge

faergeek:
See #76

### DIFF
--- a/demo/badge/badge.js
+++ b/demo/badge/badge.js
@@ -2,9 +2,20 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import Icon from '../../src/Icon';
 import Badge from '../../src/Badge';
+import Textfield from '../../src/Textfield';
 
 class Demo extends React.Component {
+    state = {
+        messages: '0'
+    }
+
+    getChangeHandler(prop) {
+        return value => this.setState({ [prop]: value });
+    }
+
     render() {
+        const { messages } = this.state;
+
         return (
             <div>
                 <p>Number badge on icon</p>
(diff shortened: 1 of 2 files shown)


src/Badge.js
```diff
@@ -25,7 +25,7 @@ class Badge extends React.Component {
 
         return React.cloneElement(element, {
             className: 'mdl-badge',
-            'data-badge': this.props.text
+            'data-badge': this.props.text || undefined
```
tleunen: why not just `this.props.text`?
Does mdl render something if it's an empty string?

demo/badge/badge.js
```diff
@@ -2,9 +2,20 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import Icon from '../../src/Icon';
 import Badge from '../../src/Badge';
+import Textfield from '../../src/Textfield';
 
 class Demo extends React.Component {
+    state = {
```
tleunen: state should be on the instance inside the constructor, not static ;)

faergeek: It isn't static, it's on instance, static should be prepended by `static` keyword :-)

src/Badge.js
faergeek: mdl has a selector that looks like this `.mdl-badge[data-badge]:after` so it renders badge if there's `data-badge` attribute even if it's empty.

And `undefined` does not render attribute at all. Check the DOM in devtools.

faergeek: We could set prop validation to `any` and convert `text` prop to string though.

faergeek: Sorry, actually we can't, undefined equals string 'undefined' if it's converted to string.

faergeek:
No need to merge commit for demo, it's just for demonstration/testing.
I'm going to remove it if it's not needed anymore.

tleunen:
So.. What you're trying to do is to not display the badge icon if the text is empty, right?

faergeek:
Exactly.

Otherwise we would need to conditionally wrap/unwrap element.

tleunen:
And React displays the attribute, even if `text` is empty?

faergeek:
Attribute is rendered until it doesn't equal to `null` or `undefined`.

tleunen:
Should be fixed now.

PR #83, opened 2015-10-17
Fix: Allow user to provide input props

### LINKED ISSUE #82: Don't change event handling
```jsx
function myHandler ( event ) {}
<input type='checkbox' onChange={ myHandler } />
```

Then, if I were to switch to use this library,
```jsx
/* <input type='checkbox' onChange={ myHandler } /> */
<Checkbox onChange={ myHandler } />
```
it breaks.

It fires twice: once with the value, once with the Event.

I would much prefer to have my event handling all work the same: handle the event, let the handler work out what they find useful.

Please change to not intercept the native events.
Same request for Textfield and others.

Thanks for the library, by the way. It's so much easier than using all the mdl classnames

joshq00:
Fixes #82

* Update change events use `event` instead of `value`.
* Props solely used for `input` components, where able, are directly passed

There are still some components where value, id, name are handled manually.

### DIFF
--- a/src/RadioGroup.js
+++ b/src/RadioGroup.js
@@ -12,7 +12,7 @@ class RadioGroup extends React.Component {
         }),
         container: PropTypes.string,
         name: PropTypes.string.isRequired,
-        onChange: PropTypes.func.isRequired,
+        onChange: PropTypes.func,
         value: PropTypes.oneOfType([
             PropTypes.string,
             PropTypes.number
@@ -23,20 +23,16 @@ class RadioGroup extends React.Component {
         container: 'div'
     }
 
-    _handleChange = (value) => {
-        this.props.onChange(value);
-    }
-
     render() {
-        var { name, onChange, value, children,
+        var { name, value, children,
             container, childContainer, ...otherProps} = this.props;
 
         return React.createElement(container, otherProps,
             React.Children.map(children, child => {
                 var clonedChild = React.cloneElement(child, {
-                    name: name,
                     checked: child.props.value === value,
-                    onChange: this._handleChange
+                    name,
+                    ...otherProps
                 });
 
                 return childContainer ? React.createElement(childContainer, {}, clonedChild) : clonedChild;
--- a/src/Slider.js
+++ b/src/Slider.js
@@ -5,18 +5,14 @@ import mdlUpgrade from './utils/mdlUpgrade';
 class Slider extends React.Component {
     static propTypes = {
         className: PropTypes.string,
-        max: PropTypes.number.isRequired,
-        min: PropTypes.number.isRequired,
-        onChange: PropTypes.func.isRequired,
+        max: PropTypes.number,
+        min: PropTypes.number,
+        onChange: PropTypes.func,
         value: PropTypes.number
     }
 
-    _handleChange = (event) => {
-        this.props.onChange(parseFloat(event.target.value));
-    }
-
     render() {
-        var { className, max, min, onChange, value, ...otherProps } = this.props;
+        var { className, ...otherProps } = this.props;
 
         var classes = classNames('mdl-slider mdl-js-slider', className);
 
@@ -24,11 +20,6 @@ class Slider extends React.Component {
             <input
                 className={classes}
                 type="range"
-                min={min}
-                max={max}
-                value={value}
-                tabIndex="0"
-                onChange={this._handleChange}
                 {...otherProps}
             />
         );
--- a/src/Tabs/Tabs.js
+++ b/src/Tabs/Tabs.js
@@ -13,7 +13,7 @@ class Tabs extends React.Component {
             }
         }),
         className: PropTypes.string,
-        onChange: PropTypes.func.isRequired,
+        onChange: PropTypes.func,
         ripple: PropTypes.bool
     }
 
@@ -22,7 +22,9 @@ class Tabs extends React.Component {
     }
 
     _handleClickTab = (tabId) => {
-        this.props.onChange(tabId);
+        if (this.props.onChange) {
+            this.props.onChange(tabId);
+        }
     }
 
     render() {
--- a/demo/slider/slider.js
+++ b/demo/slider/slider.js
@@ -3,9 +3,9 @@ import ReactDOM from 'react-dom';
 import Slider from '../../src/Slider';
 
 function linkToState(target, property) {
-    return value => {
+    return event => {
         target.setState({
-            [property]: value
+            [property]: event.target.value
         });
     };
 }
--- a/demo/textfield/textfield.js
+++ b/demo/textfield/textfield.js
@@ -3,9 +3,9 @@ import ReactDOM from 'react-dom';
 import Textfield from '../../src/Textfield';
 
 function linkToState(target, property) {
-    return value => {
+    return event => {
         target.setState({
-            [property]: value
+            [property]: event.target.value
         });
     };
 }
--- a/demo/toggle/toggle.js
+++ b/demo/toggle/toggle.js
@@ -6,18 +6,17 @@ import Radio from '../../src/Radio';
 import IconToggle from '../../src/IconToggle';
 import Switch from '../../src/Switch';
 
-function linkToState(target, property) {
-    return value => {
+function linkValueToState(target, property) {
+    return event => {
         target.setState({
-            [property]: value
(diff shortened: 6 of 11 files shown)


tleunen:
Thanks @joshq00! I really appreciate the help to make `react-mdl` better :)

I understand we could do a cleanup in the `propTypes` of every component, but is some cases and especially when they are flagged "required", I think it's good to have them. It forces the user to use/set them and to know what's available on the component.

joshq00:
I don't see any problem with keeping them in the `propTypes`, or documenting that all `input` attributes are valid props.

However, I don't think they should be required if they are not required on the native input element.

For example, if I want a text box without a label (maybe I'm just using `placeholder`, or putting my label elsewhere) - I have to do
```jsx
<Textfield label="" />
```
Without doing label="", I don't just get an invariant warning... I actually get an error thrown

tleunen:
I think it really depends. Because you know how the component is built, you know you can use a lot of props that are not described. But ideally, props should describe what you can use (or at least those who are supported and tested). A slider without the min/max value for example doesn't make a lot of sense, but because it's an input behind the scene, you know you can provide them.

joshq00:
I made a [fiddle](http://jsfiddle.net/4n5nqxya/1/) testing behavior with/without labels

tleunen:
Yes the input still works. But it doesn't have the MDL styles. Label is required with a MDL Textfield, we cannot change that.

joshq00:
Fair enough.

Anyhow, I didn't remove `isRequired` from the label property. Would you like me to add other properties?

tleunen:
For now, because there's no documentation, I prefer keeping the propTypes because they are kind of a documentation as well.

Could you rebase instead of merge? To avoid having these merge commits ? :)
Thanks!

joshq00:
Rebased

tleunen:
Could you please only change the way the event is sent inside `onChange`?
As I said, I prefer keeping the other props to serve as a documentation.

joshq00:
I'll get those back in today or tomorrow at latest

joshq00:
@tleunen should be all ready now

src/RadioGroup.js
```diff
@@ -12,7 +12,7 @@ class RadioGroup extends React.Component {
         }),
         container: PropTypes.string,
         name: PropTypes.string.isRequired,
-        onChange: PropTypes.func.isRequired,
+        onChange: PropTypes.func,
```
tleunen: any reason why it's not required anymore?
Usually, React wants us to provide an `onChange` callback if we set a `value`.
This is valable for every input.

src/Slider.js
```diff
@@ -5,18 +5,14 @@ import mdlUpgrade from './utils/mdlUpgrade';
 class Slider extends React.Component {
     static propTypes = {
         className: PropTypes.string,
-        max: PropTypes.number.isRequired,
-        min: PropTypes.number.isRequired,
-        onChange: PropTypes.func.isRequired,
+        max: PropTypes.number,
```
tleunen: Why it's not required anymore?
Creating a Slider without the min/max doesn't make any sense

src/Slider.js
```diff
@@ -5,18 +5,14 @@ import mdlUpgrade from './utils/mdlUpgrade';
...
+        min: PropTypes.number,
+        onChange: PropTypes.func,
```
tleunen: Same here, I can't think of a scenario where we won't set the callback

src/Slider.js
```diff
@@ -24,11 +20,6 @@ class Slider extends React.Component {
             <input
                 className={classes}
                 type="range"
-                min={min}
-                max={max}
-                value={value}
-                tabIndex="0"
```
tleunen: It's needed. Upstream set the value so there's no reason we don't do the same.

src/Tabs/Tabs.js
```diff
@@ -13,7 +13,7 @@ class Tabs extends React.Component {
             }
         }),
         className: PropTypes.string,
-        onChange: PropTypes.func.isRequired,
+        onChange: PropTypes.func,
```
tleunen: Same, isRequired?

src/RadioGroup.js
joshq00: Exactly my reasoning. React already handles the checking and it's not required on the true input element. Warnings will show as they should, and the user will interact with the element like they would a normal input element.

As a user, I might not care to listen to the `onChange` event. I may be prototyping.

Especially in the case of the other input types, I often use `defaultValue` instead of `value`+`onChange`

Making the fewest assumptions about what the user "meant to do" keeps the library from limiting what they can do

src/Slider.js
joshq00: But it's allowed. None of those are required properties on input elements. For example, `<input type="range" max="100" />` is valid.

Also, consider `<Slider disabled />`

src/Slider.js
joshq00: `<Slider disabled />`

src/RadioGroup.js
tleunen: Yeah the only reason why I see onChange not beeing required is just for prototyping because otherwise you have no way to get the value from the input.

src/Slider.js
joshq00: `<Slider defaultValue={ 25 } max={ 100 } />`

src/RadioGroup.js
joshq00: Or, one could have a dom listener registered

tleunen: seriously?! Either you use react, or you don't...

I understand it's nice to be generic, but I think there's a limit to that lol

joshq00: Not exactly. Take Facebook for example. The majority of the site was done directly mutating dom and they implemented react just for the chat section. There's a reason Facebook doesn't make assumptions about what we _meant_ to do. I can have a full blown site that I just drop in a React component for one piece. That shouldn't break the rest of my app.

src/Slider.js
tleunen: Indeed. It seems they have a default min-max of 0-100 with 20 as the default value.
Agreed with disabled.

src/RadioGroup.js
joshq00: Another example is if the user is looking at `onClick` instead of `onChange`. Or `onTouchStart`/`onTouchEnd`

Assumptions make an opinionated framework and, if I wanted that, I'd use ember or angular

src/Slider.js
tleunen: I was taking about the `tabIndex` ;)

tleunen:
In general, yes we can set the onChange callback as optional. It helps prototyping screen. But usually people will set it anyway because we usually create controlled form. (using value + onChange).

src/Slider.js
joshq00: Doesn't tabindex have a default of 0 anyhow? I'd have to check into it, but I'm pretty sure that's the case

joshq00:
100% agree - you'll probably will want to register a handler... but you might not. there are a million edge cases where you could be accomplishing the same thing.

A use case for not having `value` is if I'm using a store to keep track of the freeform text instead of state.
Another use case is that, sometimes, I need to use `onKeyUp` instead of `onChange`. By requiring it, I need to do `onChange={ () => {} }` just to avoid all the warnings

src/Slider.js
tleunen: Yep, I think it's the default value. Might be better to ask upstream why they specify the tabIndex value then before changing it here. Because they have it in all their demos.

joshq00: I just tried it : without `tabIndex=0`, tabbing through still works properly.

tabindex=0 should be the default for any focusable element

tleunen: Could you open an issue upstream to ask them and follow up on this?

joshq00: I just checked their [slider samples](http://www.getmdl.io/components/index.html#sliders-section) and, if you scroll down to "Example," they do show usage without setting `tabindex`

faergeek:
@tleunen Take a look at this about `onChange` https://github.com/tleunen/react-mdl/issues/55

tleunen:
Yep I forgot about this valueLink. But I agree with the changes. I'll merge once I get a confirmation for the tabindex thing. I asked them on their gitter channel.

PR #84, opened 2015-10-17
improves badge + tests

tleunen:
@faergeek could you please take a look at this?
I updated the badge component to allow null/undefined and in that case it won't render the badge.


tleunen:
I'll merge. please enter an issue if you still have issues with it.

PR #87, opened 2015-10-18
Render Badge if undefined or null

faergeek:
See #81 

That's how it supposed to work.
There's no need to check `text` prop at all, it just renders in resulting DOM or not and that's it :-)

Also previously it would throw if we pass `string` as a `children`.


tleunen:
I disagree with this.
If there's no reason to have a badge, there's no reason to have the badge css class as well.

faergeek:
Did you think about dynamic modifications to badge text? It would be much more performant to add/remove attribute than wrap/unwrap element, React can't deal with it in some smart and performant way AFAIK.

tleunen:
Yep, everything is fine. I made a small modification so a text will always have a span around him

Issue #57, opened 2015-09-28
add tests


tleunen:
We should add tests to make sure there are no regressions when changing/adding new code...

faergeek:
Which stack do you prefer? I'd like to try tape, do you have some experience with it?

tleunen:
I don't really have much experiences in this.. Would probably have to check what other React repo do in term of tests and use the same frameworks for that...

faergeek:
I'd like to help with that. Really want to have tests for this project.

nathanmarks:
@tleunen If you're happy with karma+mocha for testing I can assist in writing tests -- just checking out the lib now to potentially use in a project. If I do use it, I'll be more than happy to contribute tests.

tleunen:
@nathanmarks sure thanks!
I'm still unsure if karma/mocha is really needed of if we just have to check the output html for each components based on the props we provide.
I don't think we really need to render it in a real browser in this case, because MDL upstream already does this kind of testing to making sure the rendering is correct.

faergeek:
@tleunen It doesn't make sense to check rendered html and even check parts of a DOM, but we should test if interactions work correctly.

faergeek:
Not sure if shallow render can help with that. MDL works in a browser environment.

faergeek:
@nathanmarks Thanks, I think we need some help with tests.
Do you have any experience in writing tests with user interactions?

nathanmarks:
@faergeek Are you talking about simulating real user interactions? I write quite a bit of scripting for headless browsers at work.

faergeek:
Yes, react has test utils to simulate events https://facebook.github.io/react/docs/test-utils.html.

But I'm pretty sure that they're simulating events just for react itself and we probably need something like selenium to test the whole thing.

nathanmarks:
@faergeek I have used selenium a bit and I've got a lot of phantomjs experience. 

I think if @tleunen can pitch in and let us know what he sees as in scope for testing here -- we can make a decision on tape/mocha/etc.

I would start by asking ourselves, what is the goal of this package, and what is the best way to confirm that the goal is being reached?

If the goal is to 100% mirror the MDL upstream functionality, perhaps testing render output in different component states (via user interaction) and comparing it to the MDL output is a possibility for tests.

faergeek:
@nathanmarks This project is not the mirror of upstream functionality, it's wrapper around it.

[...]

PR #91, opened 2015-10-18
Setup testing suite

faergeek:
See #57 

Here are my suggested test suite with code coverage, watching and shallow rendering utils.
I used some of my expectations about `Badge` component as an example :-)
And tests fail. It deserves separate discussion probably.

So, what you can do with it:

1. You can run tests once with `npm test`
2. You can run watcher which will rerun on change with `npm run test:watch`
3. You can produce code coverage report with `npm run test:cov` and see the report by opening `coverage/lcov-report/index.html` in browser.
But the folder is empty if tests failed.
It's possible to see it by temporarily commenting the last two tests though.

What do you think about all of that?


faergeek:
P.S. I still would like to use http://speckjs.github.io/ but it probably will require some custom code to easily test react render output.

faergeek:
Need to fix test running as it silently fails now in Travis.

tleunen:
I agree with the coverage but I have no plan to move to tape.

faergeek:
But why? There are not much tests yet, I can covert them to tape easily.

PR #93, opened 2015-10-21
A few fixes for coverage and CI

faergeek:


package.json
```diff
@@ -67,7 +67,7 @@
   },
   "scripts": {
     "test": "mocha --compilers js:babel/register --recursive",
-    "test:cov": "babel-node $(npm bin)/isparta cover $(npm bin)/_mocha --report lcovonly -- -R spec --recursive",
+    "test:cov": "babel-node $(npm bin)/isparta cover $(npm bin)/_mocha -- -R spec --recursive",
```
tleunen: `--report lcovonly --report html` instead?
No need of the other exports

faergeek: I'd like to have json too, because I'm going to use plugin for my code editor to highlight not covered lines and it understands json only.

I don't think that it takes much time to generate it, do you?

Issue #89, opened 2015-10-18
Contributing docs


joshq00:
Add a contributing guideline

PR #107, opened 2015-11-09
adds contributing document

tleunen:
resolving #89.

@joshq00, would you please review this?
I'm bad at this so it's heavily based on the document from [h5bp/html5-boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/CONTRIBUTING.md).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,90 @@
+# Contributing to React-MDL
+
+♥ React-MDL and want to get involved?
+Thanks! There are plenty of ways you can help!
+
+Please take a moment to review this document in order to make the contribution
+process easy and effective for everyone involved.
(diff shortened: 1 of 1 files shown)


faergeek:
Take a look at this doc too, mainly about issue reporting (os, browser, reduced test case) https://github.com/BrowserSync/browser-sync/blob/master/CONTRIBUTING.md

tleunen:
Yep a section similar to that also exists in h5bp, and I didn't use it because I'm not sure it's important to know the os/browser for an issue in react-mdl. I mean, we are using the css/base js from upstream, so any specific bug would need to be reported upstream, not here.

joshq00:
I think it's a good start. Things can be modified as they come up over time

PR #109, opened 2015-11-11
fix error import modules

Graf009:
fix error import modules in index.js


tleunen:
Good catch! :/

janpieterz:
Awesome! Just figured this out myself. @tleunen when do you expect to publish this to npm?

tleunen:
I was waiting to see if there were any other quick fixes to make

PR #114, opened 2015-11-13
Add CardMedia component

durandj:
This utilizes the MDL mdl-card__media class as specified in the MDL
specification. While not strictly needed it does make for more semantic markup.

### DIFF
--- a/demo/card/card.js
+++ b/demo/card/card.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { Card, CardTitle, CardText, CardActions, CardMenu } from '../../src/Card';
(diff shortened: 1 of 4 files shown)


src/Card/CardMedia.js
```diff
@@ -0,0 +1,21 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+const CardMedia = (props) => {
```
tleunen: Sounds like a good candidate for reusing the `basicClassCreator` helper.
You can see these 2 as an example: https://github.com/tleunen/react-mdl/blob/master/src/Card/index.js#L4-L5

You can keep the tests though

tleunen:
Thanks @durandj!

src/Card/CardMedia.js
durandj: That did simplify things greatly.

PR #129, opened 2015-11-24
Fix: change value to defaultValue

cadesalaberry:
if `value` is set, it cannot be modified (visually at least). `defaultValue` seems to have the expected behavior.


tleunen:
Actually, demos should probably use `defaultValue` so we can play with it for a demo purpose, but users should use `value` and `onChange` to set the value of the textfield.

darkyen:
@cadesalaberry defaultValue should be used in case of non-controlled components. Although sometime helpful they can go out of sync.

PR #141, opened 2015-12-04
RFC: Start porting mdl template

brysgo:
This is not finished yet, but I was playing around with it and figured I'd get feedback before moving forward.
- I wasn't sure where to put everything.
- My goal is to touch ReactDOM as little as possible.
- The SVG's I tried to scrape don't work yet.


tleunen:
Thanks for the work @brysgo. I'll take a deeper look this weekend, but at first, I think it would be better to put the templates "outside" of the docs.

It's a bit more work, because it would need a separate webpack config for it, but having a page `/templates` listing all the templates and then `/templates/dashboard` would be nice.

`/templates` will be a page inside the doc (so a simple markdown page, like the homepage for example), and then `/templates/dashboard` would be a complete other "website".

What do you think?

tleunen:
hey @brysgo, I just pushed 2 examples of templates in the docs. Please take a look and feel free to adapt your PR to update the dashboard template :-)

brysgo:
@tleunen - thanks, I'll take a look

brysgo:
Going to close this since it is really old and i can't remember what it is about.

PR #147, opened 2015-12-13
Changes layout component accepted PropTypes.

skeskinen:
`PropTypes.node` accepts anything that can be rendered: numbers, strings, elements or an array containing these types.

Title and logo props are only used as react nodes so this should be perfect.

Imo this is pretty trivial change, but if you want I can add regression tests.


tleunen:
Thanks @skeskinen!

PR #156, opened 2015-12-16
Call checkValidity() on error prop change

### LINKED ISSUE #152: Textfield does not always clear is-invalid when error is set to null or undefined.
Issue exhibited here:

http://jsfiddle.net/jq0Lzgwn/

Click the error link to set an error on the Textfield and then click submit which will try to set error back to null. It succeeds in blanking out the error text but the Textfield still has the is-invalid class.

slybridges:
Add test reproducing #152 
Fixes #152


tleunen:
Awesome! Thanks @slybridges!

PR #160, opened 2015-12-17
Add tabBarProps prop to Tabs and accept Tabs of only 1 element

### LINKED ISSUE #159: Tabs: Improvement Suggestions
:wave: @tleunen 

I am working with some Tabs at the moment and would like to suggest 2 improvement suggestions. 
Happy to make PR(s) if you agree with these.

#### 1/ Allow to style `TabBar`
I would prefer my Tabs to be aligned to the left (vs center by default with `mdl`).
Hence i would need to be able pass some styles to `TabBar` from the `Tabs` component (`justifyCenter: 'left'` in my case)
Are you happy to either pass the `style` prop from `Tabs` to `TabBar` or to add an optional `tabBarStyle` to `Tabs`and pass it down (maybe better)?

#### 2/ Allow Tabs to have only one Tab
In my app, a user that is logged out sees only 1 Tab but 3 if he's logged in.
When there is only 1 Tab,  `react-mdl`throws a warning because it is expecting an `arrayOf` Tabs. Are you happy for me to tweak the PropTypes definition to suppress that warning ? I am thinking of something along the lines of `oneOfType([oneOf(...), arrayOf(...)])`

Cheers

slybridges:
Closes #159

### DIFF
--- a/docs/pages/layout.md
+++ b/docs/pages/layout.md
@@ -318,6 +318,7 @@
 | Tabs      | activeTab    | Number    | Set the active tab  | Optional, default 0 |
 | Tabs      | onChange     | Function  | Set the change callback | Optional |
 | Tabs      | ripple       | Boolean   | Applies ripples effect on tabs | Optional |
+| Tabs      | tabBarProps  | Object    | Pass Props to the TabBar Element | Optional |
 
 ### Footer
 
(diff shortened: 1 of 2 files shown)


tleunen:
Could you also add the information in the docs (configuration table)?

Thanks :)

slybridges:
Not sure you get a notif when there is a new @tleunen, but done ;-)

slybridges:
hello @tleunen - could you maybe publish a new version of `react-mdl` including this PR ? My site doesn't look that pretty without it :) Thanks !

tleunen:
Released as `v1.1.0`. Sorry for the delay @slybridges

slybridges:
Top, merci Tommy ;-)

PR #161, opened 2015-12-17
Fix src path in DocApp

sarink:
When running `npm serve-docs`, this error is present:

```
ERROR in ./docs/src/DocApp.js
Module not found: Error: Cannot resolve directory '../..' in /home/vagrant/react-mdl/docs/src
 @ ./docs/src/DocApp.js 19:8-25
```


tleunen:
Thanks @sarink! I usually have the compiled files so I haven't noticed the error before :)

PR #179, opened 2016-01-06
Allow Textfield error to be of node type

Gregoor:
In case you want to display a link in the error message.

### DIFF
--- a/src/Textfield.js
+++ b/src/Textfield.js
@@ -7,7 +7,7 @@ class Textfield extends React.Component {
     static propTypes = {
         className: PropTypes.string,
         disabled: PropTypes.bool,
-        error: PropTypes.string,
+        error: PropTypes.node,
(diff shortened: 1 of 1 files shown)


tleunen:
`PropTypes.node` seems to be the right thing to use then to allow all sort of "rendering" elements.

Gregoor:
Oh, I did not know this type existed. thanks, will change!

tleunen:
From the doc
> // Anything that can be rendered: numbers, strings, elements or an array
    // (or fragment) containing these types.
    optionalNode: React.PropTypes.node,

So only `node` is enough. Also, could you please squash your commits please?

Gregoor:
It took a while but I think I squashed it. Always grateful for a chance to work on my git skills.

tleunen:
perfect! :) Thanks @Gregoor

PR #181, opened 2016-01-06
RadioGroup is now Mutable by default

Kennyomg:
Changed Checked to defaultChecked to make RadioGroup Mutable again.

The error that is now fixed:
Warning: Failed form propType: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `Radio`.


tleunen:
So should this be `checked` if there's an `onChange` callback, and `defaultChecked` if not?

tleunen:
Thanks for the PR, I fixed it by handling both controlled and uncontrolled radio groups.

PR #207, opened 2016-02-06
Add Dialog components

### LINKED ISSUE #200: Add Dialog component
With the MDL 1.1 release, a new dialog component is available: https://getmdl.io/components/index.html#dialog-section

Permagate:
Closes #200, It works with `material.js` and `material.css` in latest mdl release and the one you updated 2 hours ago (as of this comment).

Also, I have no idea how to make showDialog native method runs while testing reactDOM (I got an error with few documentations), so I have to leverage on spying the method calls themselves, if you don't mind.

Thanks!

### DIFF
--- a/docs/pages/dialogs.md
+++ b/docs/pages/dialogs.md
@@ -1,0 +1,95 @@
+# Dialogs
+
+This component uses [dialog element](https://www.w3.org/TR/2013/CR-html5-20130806/interactive-elements.html#the-dialog-element), which is only supported by Chrome and Opera currently. For other browsers, you need to include a [polyfill](https://github.com/GoogleChrome/dialog-polyfill) in your code.
+
+## Demo
+
+### Simple Dialog
+
+```jsx_demo_class
+class Demo extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  openDialog() {
+    this.setState({
+      openDialog: true
+    });
+  }
+
+  closeDialog() {
+    this.setState({
+      openDialog: false
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <Button colored onClick={this.openDialog.bind(this)} raised ripple>Show Dialog</Button>
+        <Dialog open={this.state.openDialog}>
+          <DialogTitle>Allow data collection?</DialogTitle>
+          <DialogContent>
+            <p>Allowing us to collect data will let us get you the information you want faster.</p>
+          </DialogContent>
+          <DialogActions>
+            <Button type='button'>Agree</Button>
+            <Button type='button' onClick={this.closeDialog.bind(this)}>Disagree</Button>
+          </DialogActions>
+        </Dialog>
+      </div>
+    );
+  }
+}
+```
+
+### Dialog with Full Width Actions
+
+```jsx_demo_class
+class Demo extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  openDialog() {
+    this.setState({
+      openDialog: true
+    });
+  }
+
+  closeDialog() {
+    this.setState({
+      openDialog: false
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <Button colored onClick={this.openDialog.bind(this)} raised ripple>Show Modal</Button>
+        <Dialog open={this.state.openDialog}>
+          <DialogTitle>Allow this site to collect usage data to improve your experience?</DialogTitle>
+          <DialogContent>
+            <p>Allowing us to collect data will let us get you the information you want faster.</p>
+          </DialogContent>
+          <DialogActions fullWidth>
+            <Button type='button'>Agree</Button>
+            <Button type='button' onClick={this.closeDialog.bind(this)}>Disagree</Button>
+          </DialogActions>
+        </Dialog>
+      </div>
+    );
+  }
+}
+```
+
+## Configuration
+
+| Element   | Prop         | Type      | Effect       | Remarks      |
+|:----------|:-------------|:----------|:-------------|:-------------|
+| Dialog    | open         | Boolean   | Set the open state of the dialog  | Optional |
+| DialogTitle | component  | String, Element, Function | Specify the custom component to use to render the element | Optional. Default 'h4' |
+| DialogActions | fullWidth | Boolean  | Apply the full-width effect to all children of dialog actions  | Optional |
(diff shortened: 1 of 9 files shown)


src/Dialog/Dialog.js
```diff
@@ -0,0 +1,41 @@
...
+import classNames from 'classnames';
+
+class Dialog extends React.Component {
+    static propTypes = {
+        className: PropTypes.string,
+        isOpen: PropTypes.bool
+    };
+
+    componentDidMount() {
+        if(!!this.props.isOpen) {
```
tleunen: no need of `!!`. The proptype already specify the value should be a boolean. If the user provide something else, he will be notified that he has to change it.

src/Dialog/Dialog.js
```diff
@@ -0,0 +1,41 @@
...
+            findDOMNode(this).showModal();
+        }
+    }
+
+    componentDidUpdate(prevProps) {
+        if(this.props.isOpen !== prevProps.isOpen) {
+            if(!!this.props.isOpen) {
```
tleunen: same here, no need of `!!`

src/Dialog/DialogActions.js
```diff
@@ -0,0 +1,25 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+class DialogActions extends React.Component {
```
tleunen: could be replace with a stateless function :)

src/Dialog/DialogTitle.js
```diff
@@ -0,0 +1,22 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+class CardTitle extends React.Component {
```
tleunen: Stateless function component instead?

src/Dialog/DialogTitle.js
```diff
@@ -0,0 +1,22 @@
...
+    };
+
+    render() {
+        const { className, children, ...otherProps } = this.props;
+
+        const classes = classNames('mdl-dialog__title', className);
+
+        const containerProps = { className: classes, ...otherProps };
+
+        return typeof children === 'string'
```
tleunen: `React.createElement(typeof children === 'string' ? 'h4' : 'div', containerProps, children)` ? :)

tleunen:
Awesome work @Permagate! Just minor comments about the code.
For the tests, spying is fine.

Should the dialog be automatically displayed when added in the dom, or always in the dom and use the `isOpen` prop?
I mean... It's pretty much the same between, except in one case, the element is not in the dom, unless it's really needed.
```jsx
{{shouldOpenDialog && <Dialog />}}
// and
{{<Dialog isOpen={shouldOpenDialog} />}}
```

tleunen:
nvm about the open thing. I guess it will feel more natural for users to use a specific prop, since the native way of opening a dialog is to call a `showModal` function.

Anyway, I'll check your branch locally and verify with the MDL doc to see if everything is ok :) Thanks again!

Permagate:
Hi @tleunen, thanks for the review :+1: I'll try to update the docs as well sometimes tonight.

`React.createElement(typeof children === 'string' ? 'h4' : 'div', containerProps, children)`

Regarding DialogTitle, I'm still not sure how to approach this correctly. In the [mdl sample code](http://codepen.io/Garbee/full/EPoaMj/), it appears they intend the 'mdl-dialog__title' to be used in a title element.

`<h4 class='mdl-dialog__title'>My title</h4>`

But on my tests, it can be a div as well so users can put inside whatever they want. So that's why I currently implement it like that:

* If the children is only a string, then it will be rendered as h4 element.
```
<DialogTitle>My custom title!</DialogTitle>
<!-- will become <h4 class='mdl-dialog__title'>My custom title!</h4>
```

* If the children is not a simple string, then it will be rendered as div element.
```
<DialogTitle>
    <h3>My custom title!</h3>
    <h5>My custom subtitle!</h5>
</DialogTitle>
<!-- will become
    <div class='mdl-dialog__title'>
        <h3>My custom title!</h3>
        <h5>My custom subtitle!</h5>
    </div>
-->
```

Maybe I should stick to an element?

tleunen:
The title is actually just this:
```css
.mdl-dialog__title {
    padding: 24px 24px 0;
    margin: 0;
    font-size: 2.5rem;
}
```
So basically, it could be anything. The doc uses `h4`, but the demo uses `h3` for example. Maybe we could use the same way as some other component, and having a prop `component` allowing to specify the element to use, with a default of `h4` if you want :)

src/Dialog/Dialog.js
```diff
@@ -0,0 +1,41 @@
+import React, { PropTypes } from 'react';
+import { findDOMNode } from 'react-dom';
+import classNames from 'classnames';
+
+class Dialog extends React.Component {
+    static propTypes = {
+        className: PropTypes.string,
+        isOpen: PropTypes.bool
```
tleunen: Most of the prop we use in react-mdl doesn't use the prefix `is`. Do you think it would be better to only use `open`?
The native dialog also use an `open` attribute.

Permagate: Yeah, I feel that would be better. It is inconsistent with the other component's style. I initially used isOpen because I thought I should make it different with the native open.

Give me a moment, I'll update the PR.

Permagate:
Documentation page added. Not sure if I should add a preface talking about the polyfilling dialog (which currently only works with Chrome). But I left it text-less like the other pages. Let me know if you prefer otherwise.

tleunen:
Yes adding a note about the polyfill will be helpful since the user has to include it :)

Thanks again for all this! I know there's not really any documentation to explain how to write and test the doc.

- Could you also please squash the commits into 1 instead (or several but 1 for the code, 1 test and 1 documentation)?
- Don't forget to add the Dialog in `src/index.js` as well :)

docs/pages/dialogs.md
```diff
@@ -0,0 +1,95 @@
...
+  closeDialog() {
+    this.setState({
+      openDialog: false
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <Button colored onClick={this.openDialog.bind(this)} raised ripple>Show Dialog</Button>
```
tleunen: Please rename the function `handleOpenDialog` and bind the function inside the constructor instead.
`this.handleOpenDialog = this.handleOpenDialog.bind(this);`
Same for the close event, in both examples :)

Issue #199, opened 2016-02-01
Add Snackbar component


tleunen:
With the MDL 1.1 release, a new snackbar component is available: https://getmdl.io/components/index.html#snackbar-section

PR #208, opened 2016-02-07
adds Snackbar component

tleunen:
Here's a first draft of a Snackbar (#199).

I decided to not use the JS implementation done by the MDL team since the implementation is not very React friendly.

Of course, the queue system available in the MDL implementation is not included, and will not be included.
In an application, it doesn't make sense to have multiple snackbar (they will overlap), so you should manually create a queue system to manage all type of notifications.

Here's the example with the button color change: http://codepen.io/anon/pen/GoYzLV?editors=0010

### DIFF
--- a/src/Snackbar.js
+++ b/src/Snackbar.js
@@ -1,0 +1,78 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+// This component doesn't use the javascript from MDL.
+// This is the expected behavior and the reason is because it's not written in
+// a way to make it easy to use with React.
+const ANIMATION_LENGTH = 250;
+
+class Snackbar extends React.Component {
+    static propTypes = {
+        action: PropTypes.string,
+        active: PropTypes.bool.isRequired,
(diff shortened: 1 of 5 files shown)


src/Snackbar.js
```diff
@@ -0,0 +1,75 @@
...
+// This is the expected behavior and the reason is because it's not written in
+// a way to make it easy to use with React.
+const ANIMATION_LENGTH = 250;
+
+class Snackbar extends React.Component {
+    static propTypes = {
+        action: PropTypes.string,
+        active: PropTypes.bool.isRequired,
+        className: PropTypes.string,
+        onClick: PropTypes.func,
```
Permagate: Personally, when I saw onClick, I thought it was a handler when snackbar is clicked (not just the action button). I think we should keep mdl naming (actionHandler). Or if we want to stick to `on-` naming, maybe onActionClick?

tleunen: hmm I'll rename it with `onActionClick` then :)

Issue #201, opened 2016-02-01
Add List component


tleunen:
With the MDL 1.1 release, a new list component is available: https://getmdl.io/components/index.html#lists-section

darenju:
I have started working on that. Should I create my PR?

tleunen:
Yep you can create a PR and I'll review it :)
Thanks a lot!

PR #212, opened 2016-02-10
Adds support of List

darenju:
Adds support of List, and List Item component. #201

Still thinking of how "primary content", "secondary content" & friends should be implemented. Would need help on that.


tleunen:
I'd probably go with a specific primary content, and a secondary content to have everything flexible.
I'll take a deeper look later today and read the spec and mdl doc to see how we could do it.

darenju:
So, we'd have a PrimaryContent and SecondaryContent components? Also, there's this difference between "secondary content" and "secondary actions", and the latter can be a link or a button, if I'm not mistaken.

tleunen:
The `ListPrimaryContent` could take an `icon` and a `subtitle` as props, and `children` will be the title.
The `ListSecondaryContent` could take an `info` as prop and `children` will be the action.

I'm just throwing some ideas, I'd need to think about it. I'd like to provide an easy way to create a list of items (where the item is a link, like in the drawer for example), and the possibility to have that fully customizable :)

darenju:
The `ListPrimaryContent` and `ListSecondaryContent` seem pretty decent.

However, I don't get your second paragraph. What we could do is providing a `ListItemAction` that would take a `icon`, a `onClick` and a child content for label. What do you think about this?

tleunen:
Do you mean having List > ListItem > ListItemPrimary & ListItemSecondary > ListAction ? Or replacing Secondary with Action?
Could you write an example of jsx code? It will help me visualize it

darenju:
For example, this code (3rd example on Components section of MDL):
```html
<div class="demo-list-action mdl-list">
  <div class="mdl-list__item">
    <span class="mdl-list__item-primary-content">
      <i class="material-icons mdl-list__item-avatar">person</i>
      <span>Bryan Cranston</span>
    </span>
    <a class="mdl-list__item-secondary-action" href="#"><i class="material-icons">star</i></a>
  </div>
  <div class="mdl-list__item">
    <span class="mdl-list__item-primary-content">
      <i class="material-icons mdl-list__item-avatar">person</i>
      <span>Aaron Paul</span>
    </span>
    <a class="mdl-list__item-secondary-action" href="#"><i class="material-icons">star</i></a>
  </div>
  <div class="mdl-list__item">
    <span class="mdl-list__item-primary-content">
      <i class="material-icons mdl-list__item-avatar">person</i>
      <span>Bob Odenkirk</span>
    </span>
    <span class="mdl-list__item-secondary-content">
      <a class="mdl-list__item-secondary-action" href="#"><i class="material-icons">star</i></a>
  </span>
  </div>
</div>
```

Would be:
```jsx
<List>
  <ListItem>
    <ListItemPrimaryContent icon="person">Bryan Cranston</ListItemPrimary>
    <ListItemAction icon="star" onClick={/* ... */} />
  </ListItem>
  {/* ... */}
</List>
```

tleunen:
Oh I see... The secondary content is always an action... Well.. the [spec](https://www.google.com/design/spec/components/lists.html#lists-content) calls everything an action, but MDL only use action for the secondary content.

I won't put any event handler on that item, because it can be a checkbox or any other component...
So it could be `<ListItemAction info="Actor"><Checkbox /></ListItemAction>`

tleunen:
I think we could rename everything though.. `List`, `ListItem`, `ListItemContent` and `ListItemAction`

I'm suggesting this:
`List` and `ListItem` are basic.
`ListItemContent` would be the primary content, can receive `icon` and `subtitle` as props. Subtitle is only used for 2 and 3 lines content, and for 3 lines, subtitle will use the body specific className, and `children` will be the main content.
`ListItemAction` is the secondary content, can take `info` as props (to display a small "header"), and `children` will be its content. I think it should also attach the className `mdl-list__item-secondary-action` on the child.

What do you think?

darenju:
Everything seems clearer and precise now. Do we still agree that `ListItem` has `twoLine` and `threeLine` prop?

If so, then it's alright. Attaching `mdl-list__item--secondary-action` to the child rather than on the parent seems actually more compliant to the specs.

If everything is alright, I could start writing it today.

darenju:
I have read your comment and realised my question isn't relevant anymore. Since we would rely on presence of `subtitle` to know if the `ListItem` has two or three lines, then we do not need the `twoLine` and `threeLine` props anymore. Is that correct?

tleunen:
We can see this of 2 different ways I thing. Either the component takes a `twoLines` or `threeLines` prop. (or a single prop "lines" which takes 2 or 3?), or based on a `subtitle` or `body` prop, we can define if it's 2 or 3 lines. (subtitle = 2; body = 3).

I don't know if using `subtitle` and `body` is really good, since it can create confusion for the users. Using the number of lines is clear and you know exactly what you'll get.

darenju:
Keeping the `twoLine` and `threeLine` (with trailing S or not?) is better than specifying `body` or `subtitle` as you mentioned it.

However, the point with CSS classes is that the *parent* must have it (`-two-line` or `-three-line`). Actually I found a way to pass this prop to `ListItemContent` if it's defined in `ListItem`.

I think we're all set, aren't we?

tleunen:
Yep :-) Lets see some code and we'll see how it behaves :)

darenju:
I'm currently working on it.

Have actually done a lot, and the trickiest part is over, I guess. Expect more commits by the end of the day. :)

darenju:
I can't get to pass the Travis tests while it passes on my machine... Do you have any idea what it could be? @tleunen

tleunen:
You can see the job task here https://travis-ci.org/tleunen/react-mdl/builds/108559317
Without investigating, I have no idea, but I think your tests don't need `renderDOM`, you could only use `render` instead and check the children with `output.props.children`

tleunen:
You can see the job task here https://travis-ci.org/tleunen/react-mdl/builds/108559317
Without investigating, I have no idea, but I think your tests don't need `renderDOM`, you could only use `render` instead and check the children with `output.props.children`

Please come on discord if you wanna discuss about it ;)

src/List/ListItemAction.js
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    render() {
+        const { className, info } = this.props;
+
+        const classes = classNames('mdl-list__item-secondary-content', className);
+
+        const children = React.Children.map(this.props.children, child => {
+            const toAdd = 'mdl-list__item-secondary-action';
+            const childClassName = child.props.className ? `${child.props.className} ${toAdd}` : toAdd;
```
tleunen: you can use `classNames(child.props.className, toAdd)` instead :)

src/List/ListItem.js
```diff
@@ -0,0 +1,46 @@
...
+    static defaultProps = {
+        twoLine: false,
+        threeLine: false
+    };
+
+    render() {
+        const { className, twoLine, threeLine, ...otherProps } = this.props;
+
+        const classes = classNames('mdl-list__item', {
+            'mdl-list__item--two-line': twoLine,
```
tleunen: `twoLine && !threeLine`.
same on the line below

(I'm still unsure about using a trailing `s` or not. Feels weird to not have it though :/)

src/List/ListItem.js
```diff
@@ -0,0 +1,46 @@
...
+        }, className);
+
+        const children = React.Children.map(otherProps.children, child => {
+            let component = child;
+
+            if (child.type === ListItemContent) {
+                // Pass the threeLine prop in order to define the correct className
+                // later in ListItemContent.
+
+                component = <ListItemContent {...child.props} threeLine />;
```
tleunen: Please use `cloneElement` instead. Also, I think it would be better to use another prop name instead of `threeLine` in this case. `useBodyClass` maybe... (Anyway it's just something internal, but at least it gives more info for the `ListItemComponent`.
```
React.cloneElement(child, {
    ...child.props,
    useBodyClass: threeLine
});
```

src/List/ListItemAction.js
```diff
@@ -0,0 +1,47 @@
...
+
+class ListItemAction extends React.Component {
+    static propTypes = {
+        children: PropTypes.element.isRequired,
+        className: PropTypes.string,
+        info: PropTypes.string
+    };
+
+    static defaultProps = {
+        info: null
```
tleunen: not required

src/List/ListItemAction.js
```diff
@@ -0,0 +1,47 @@
...
+
+        let infoElement = null;
+
+        if (info) {
+            infoElement = <span className="mdl-list__item-secondary-info">{info}</span>;
+        }
+
+        return (
+            <span className={classes}>
+                {infoElement}
```
tleunen: use `{info && <span ...>}`, it's easier to read and to see what it renders :)

src/List/ListItemAction.js
```diff
@@ -0,0 +1,47 @@
...
+    };
+
+    render() {
+        const { className, info } = this.props;
+
+        const classes = classNames('mdl-list__item-secondary-content', className);
+
+        let { children } = this.props;
+
+        try {
```
tleunen: I'm not a big fan of try catch, especially in this case. The propType should be enough.

src/List/ListItemAction.js
```diff
@@ -0,0 +1,47 @@
...
+            const child = Children.only(children);
+            children = cloneElement(child, {
```
tleunen: `const child = cloneElement(Children.only(children), {...`
No need to override children.

src/List/ListItemAction.js
```diff
@@ -0,0 +1,47 @@
...
+        let infoElement = null;
+
+        if (info) {
+            infoElement = <span className="mdl-list__item-secondary-info">{info}</span>;
+        }
+
+        return (
+            <span className={classes}>
+                {infoElement}
+                {children}
```
tleunen: `{child}` base on the comment above on `const child = ...`

src/List/ListItemAction.js
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        let infoElement = null;
+
+        if (info) {
+            infoElement = <span className="mdl-list__item-secondary-info">{info}</span>;
+        }
+
+        return (
+            <span className={classes}>
```
tleunen: `<span className={classes} {...otherProps}>`

src/List/ListItemContent.js
```diff
@@ -0,0 +1,58 @@
...
+        icon: PropTypes.string,
+        subtitle: PropTypes.oneOfType([
+            PropTypes.element,
+            PropTypes.string
+        ]),
+        threeLine: PropTypes.bool
+    };
+
+    static defaultProps = {
+        icon: null,
```
tleunen: Not needed. Having `undefined` as default is enough

src/List/ListItemContent.js
```diff
@@ -0,0 +1,58 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+class ListItemContent extends React.Component {
+    static propTypes = {
+        className: PropTypes.string,
+        icon: PropTypes.string,
+        subtitle: PropTypes.oneOfType([
```
tleunen: `PropTypes.node` instead, i's for "anything that can be rendered"

src/List/ListItemContent.js
```diff
@@ -0,0 +1,58 @@
...
+            subtitleElement = (
+                <span className={subtitleClassName}>
+                    {subtitle}
+                </span>
+            );
+        }
+
+        return (
+            <span className={classes}>
+                {iconElement}
```
tleunen: use the format `{icon && <Icon name={icon} />}` (with the Icon component)
btw, we have to support `mdl-list__item-icon` and `mdl-list__item-avatar`.

Might be better to use a specific component for that as well, to allow users to set something else than a Material Icon (an image for example).

darenju:
Thank you for your feedback! It helps me get better with React.

I have fixed most of the things, except the part for the icon/avatar point. We need to design this as well.

src/List/List.js
```diff
@@ -0,0 +1,3 @@
+import basicClassCreator from '../utils/basicClassCreator';
```
tleunen: You can move this in a `src/List/index.js` with the other imports as well. No need of a specific file since the end user will never import only this one (he won't be able to do much only with this)

src/List/ListItem.js
```diff
@@ -0,0 +1,49 @@
...
+        }, className);
+
+        const children = Children.map(otherProps.children, child => {
+            let component = child;
+
+            if (child.type === ListItemContent) {
+                // Pass the threeLine prop in order to define the correct className
+                // later in ListItemContent.
+
+                component = cloneElement(child, {
```
tleunen: `return cloneElement(...`? Then we can remove the `let component = child` :)

src/List/__tests__/List-test.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-env mocha */
```
tleunen: this test can be removed. The list is just a component from the basicClassCreator, which already have tests.

src/List/ListItemContent.js
```diff
@@ -0,0 +1,38 @@
...
+    render() {
+        const { children, className, icon, subtitle, useBodyClass, ...otherProps } = this.props;
+
+        const classes = classNames('mdl-list__item-primary-content', className);
+
+        const subtitleClassName = useBodyClass ? 'mdl-list__item-text-body' : 'mdl-list__item-subtitle';
+
+        return (
+            <span className={classes}>
+                {icon && <Icon name={icon} />}
```
tleunen: Hmm... I don't really want to create a specific component for that one, so I'll try something here...

Have `icon` and `avatar` as prop. (Both are exclusive, we cannot have an icon, and an avatar at the same time.
Both can take either a string (icon name), or an element. If it's a string, we create an Icon and attach the specific classname. If it's an element, we clone it and attach the classname as well.

src/List/ListItem.js
darenju: I don't agree on this one, as for testability it's better to return only one value.

darenju:
Okay, tomorrow I'll work on `avatar` and everything else. Thanks for your feedback again.

src/List/ListItem.js
tleunen: for testability?
```js
const children = Children.map(otherProps.children, child => {
    if(child.type === ListItemContent) {
        return cloneElement(child, {
            ...child.props,
            useBodyClass: threeLine
        });
    }
    return child;
});
```

darenju: We would need to test two cases here: the regular case and the case where it's a `ListItemContent` two `return`s would mean adding another test case. At least that's what I learned. And, I previously had this same exact code, but code coverage was lower (due to what I just explained.)

tleunen: Even if you have 1 return, you still have to cover both cases. But maybe I missing something ;)

darenju: Okay then, I will rollback to two `return` statements and will add tests.

tleunen:
Any news on this? I can continue the work if needed

tleunen:
Thank you @darenju for the work. I really would like to publish the new version today so I'll complete the work here with the said modifications and the documentation.

darenju:
Sorry I haven't replied. I actually did some work, but was celebrating my birthday meanwhile. I'm glad my work has been merged. :+1:

PR #213, opened 2016-02-11
Removes key prop on inner components of Textfield

### LINKED ISSUE #197: Textfield that replaces another Textfield doesn't render properly
I encontered an issue when rendering a component that dynamically decides which fields are visible.

Here's a small example that exhibits the issue. It's a form with two fields, only one of which is visible at a given time. To keep things simple, I put a switch to select which field is visible. In my actual use case, this depends on the values of previous fields.

```javascript
import {Switch, Textfield, Button} from 'react-mdl';

const TestComponent = React.createClass({

    getInitialState() {
        return {
            showFirst: true,
        }
    },

    onChangeSelector(event) {
        this.setState({showFirst: !this.state.showFirst});
    },

    onChangeFirst(event) {
        this.setState({first: event.target.value});
    },

    onChangeSecond(event) {
        this.setState({second: event.target.value});
    },

    render() {
        return (
            <div>
                <Switch checked={this.state.showFirst} onChange={this.onChangeSelector}>Show first field</Switch>
                {
                    this.state.showFirst
                    ? <Textfield label="first" value={this.state.first} onChange={this.onChangeFirst} />
                    : <Textfield label="second" value={this.state.second} onChange={this.onChangeSecond} />
                }
                <pre>state = {JSON.stringify(this.state)}</pre>
            </div>
        )
    },

});
```

Initially the first field is visible. After typing "abc" in the first field, I get this, which looks good:

screen shot 2016-01-30 at 19 23 58

Then I flick the switch. After typing "def" in the second field, I get this, which looks less good:

screen shot 2016-01-30 at 19 24 11
 
I believe this is the symptom of a desynchronization between the React and MDL components lifecycles. I'll see if I can track this down further.

tleunen:
This will fix #197 (cc @aaugustin)

The key prop on the input was the cause of the issue. Because the key changed, React was removing the element, and adding a new one. But because the whole textfield was never removed from the DOM, the function `componentWillUnmount` on the `MDLComponent` was never called.

This codepen shows the fix in action: http://codepen.io/anon/pen/EPRLww

### DIFF
--- a/src/Textfield.js
+++ b/src/Textfield.js
@@ -11,6 +11,7 @@ class Textfield extends React.Component {
         expandable: PropTypes.bool,
         expandableIcon: PropTypes.string,
         floatingLabel: PropTypes.bool,
+        id: PropTypes.string,
         inputClassName: PropTypes.string,
         label: PropTypes.string.isRequired,
         maxRows: PropTypes.number,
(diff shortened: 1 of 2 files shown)


aaugustin:
Thanks a lot for working on this!

I tried running a version with this change in my development environment to check that it fixes my original issue. Unfortunately I failed, due to my lack of knowledge of npm.

My latest attempt was:

```
cd node_modules
rm -r react-mdl
git clone https://github.com/tleunen/react-mdl.git
cd react-mdl
git checkout 1c926e9a5835349fc318667070b697704b019cb7
npm install
npm run compile
```

At this point, my app fails with:

> Error: addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).

This must be because `node_modules` and `node_modules/react-mdl/node_modules` each have their copy of React.

tleunen:
Based on what I found on stackoverflow, it seems to be related to multiple React versions, as you suggest. It might be because the version you're getting uses React 0.14.7, and you're using something else. If you're using Webpack, you can try to add this [plugin to remove duplication](https://github.com/webpack/docs/wiki/optimization#deduplication).
If it doesn't work, one other hack could be to compile the file, copy it somewhere, reinstall the default npm version, and put back the saved compiled Textfield.js ;)

Note that instead of getting the commit, you can also use the branch name `issue-197`

aaugustin:
Indeed, I could run the new code by compiling react-mdl and replacing the node_modules/react-mdl/lib by the compiled output.

This patch clearly goes in the right direction.

Truth be said, I could still break your latest codepen as follows:

- open http://codepen.io/anon/pen/EPRLww
- click in first field
- type "foo"
- click on toggle
- click on second field
- type "bar"
- click on toggle
- click on first field
- select all text and remove it
- click on toggle

At this point, the "bar" text is superimposed on the "second" label.

Admittedly, it's an edge case. I'm not sure how much effort it is reasonable to put into fixing it!

tleunen:
@aaugustin This is because `findDOMNode` was not defined in the code of the pen.
I fixed it by adding `var findDOMNode = ReactDOM.findDOMNode;` and I can't reproduce your issue ;)
http://codepen.io/anon/pen/EPrywo

aaugustin:
I guess we're good, then. Thanks a lot for your investigations on this issue!

tleunen:
np, thanks! :)

PR #214, opened 2016-02-12
adds position prop on Tooltip

### LINKED ISSUE #211: Tooltip positioning
Here on the [mdl docs](http://www.getmdl.io/components/index.html#tooltips-section) for tooltips, near the bottom they mention 4 different classes that can be applied to a tooltip that allow you to define its open position. For everyone's convenience:

MDL class	| Effect	| Remarks 
--------------------|------------|--------------
mdl-tooltip--left	  |      Positions the tooltip to the left of the target |	Optional; goes on tooltip container element
mdl-tooltip--right |	Positions the tooltip to the right of the target |	Optional; goes on tooltip container element
mdl-tooltip--top |	Positions the tooltip to the top of the target |	Optional; goes on tooltip container element
mdl-tooltip--bottom	| Positions the tooltip to the bottom of the target |	Optional; goes on tooltip container element
```

Thanks!

### PR DESCRIPTION
**Pull request by Permagate**
Closes #211. A pretty quick change, so let me know if I miss anything.

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -3,7 +3,7 @@ import classNames from 'classnames';
 import MDLComponent from './utils/MDLComponent';
 
 const Tooltip = (props) => {
-    const { label, large, children, ...otherProps } = props;
+    const { label, large, children, position, ...otherProps } = props;
     const id = Math.random().toString(36).substr(2);
 
     const newLabel = (typeof label === 'string')
@@ -25,7 +25,8 @@ const Tooltip = (props) => {
                 {React.cloneElement(newLabel, {
                     htmlFor: id,
                     className: classNames('mdl-tooltip', {
-                        'mdl-tooltip--large': large
+                        'mdl-tooltip--large': large,
+                        [`mdl-tooltip--${position}`]: typeof position !== 'undefined'
                     })
                 })}
             </MDLComponent>
@@ -36,7 +37,8 @@ const Tooltip = (props) => {
 Tooltip.propTypes = {
     children: PropTypes.node.isRequired,
     label: PropTypes.node.isRequired,
-    large: PropTypes.bool
+    large: PropTypes.bool,
+    position: PropTypes.oneOf(['left', 'right', 'top', 'bottom'])
 };
 
 export default Tooltip;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Tooltip.js [hunk 1]
```diff
@@ -25,7 +25,8 @@ const Tooltip = (props) => {
                 {React.cloneElement(newLabel, {
                     htmlFor: id,
                     className: classNames('mdl-tooltip', {
-                        'mdl-tooltip--large': large
+                        'mdl-tooltip--large': large,
+                        [`mdl-tooltip--${position}`]: true
```
tleunen: should use `typeof position !== 'undefined'`, because we don't want a useless css class called `mdl-tooltip--undefined` when the position is not specified ;-)

src/__tests__/Tooltip-test.js
```diff
@@ -86,4 +86,90 @@ describe('Tooltip', () => {
...
+        );
+        const output = render(element);
+
+        const label = output.props.children[1].props.children;
+
+        expect(label.props.className)
+            .toInclude('mdl-tooltip--bottom');
+    });
+
+    it('should display console warning if specified position is invalid', () => {
```
tleunen: Not sure this test is really necessary. The user will see the warning in the console, that's enough ;)

For the other positions, you could only set one test and test several things inside by looping through the position. `['left', 'right', 'top', 'bottom'].forEach(() => { test... });`

src/Tooltip.js
Permagate: You are right, I should have checked my demo more thoroughly.

src/__tests__/Tooltip-test.js
Permagate: Oh yeah, I'm a bit confused about testing warning in the console. I saw similar scenarios in other components (for example [Tabs-test.js:85](https://github.com/tleunen/react-mdl/blob/master/src/Tabs/__tests__/Tabs-test.js#L85), but it simply render the element without expecting anything. I'll just remove it then.

tleunen: Hmm.. Yep, but this one is a little bit different, because it checks the proptype. I should add a spy on it though, and see if I can get the `new Error` value without spying the console

Issue #232, opened 2016-02-23
Menu broke after changing target key from htmlFor to data-mdl-for


broveloper:
Menu component no longer works with IconButton with the change from htmlFor to data-mdl-for

https://github.com/tleunen/react-mdl/commit/a4211e61f80ea78525939f4ea72680b7e4dd4fd2#diff-e359c93bc52c848fbef7e3bdc25527db

Suggestion:
Switch it back or temporarily have both?

broveloper:
temporary solution for anyone who is experiencing is to pass in the htmlFor={target} in the <Menu> component itself.

broveloper:
Not an issue.  Bug was because of using old version of MDL v1.0.6, instead of v1.1.1

PR #233, opened 2016-02-23
Hotfix for issue (fix for #232)

broveloper:
https://github.com/tleunen/react-mdl/issues/232

Add back the htmlFor property.  Didn't remove the data-mdl-for property because not sure if it would affect any other things.


broveloper:
Not an issue.  Bug was because of using old version of MDL v1.0.6, instead of v1.1.1

Issue #221, opened 2016-02-16
Dialog state does not change on close


t1bb4r:
I think the Dialog component should set its state to `{openDialog: false}` on 'dialogclose'.

Currently if a user presses the escape key to close the dialog it wont open again since the openDialog state is still set to true.
This can be reproduced using the demo page:
 - clicking show modal opens the dialog
 - press escape
 - clicking show modal again does not open the dialog

Unexpected state can easily be inspected using the react developer tools.

Edit: I'm using Chrome Version 48.0.2564.82 (64-bit)

t1bb4r:
I'm very new to React so forgive my ignorance, here is what I've done to solve it. Hope this helps some one.

The following did not work (The event might be unsupported by React?)
```
<Dialog open={this.state.openDialog} onCancel={this.handleCloseDialog}>
```
So I did the following:
```
componentDidMount() {
    ReactDOM.findDOMNode(this.refs.dialog).addEventListener("cancel", this.handleCloseDialog);
}
render() {
  ...
     <Dialog open={this.state.openDialog} ref='dialog'>
  ...
}
```
Now when a user presses escape, or exits the dialog without using the button, the state is correctly set and the dialog can be opened once more.

tleunen:
Thanks for reporting the issue!
I didn't know the escape key could cancel the dialog. But indeed, it's explained on the [Chrome developer page](https://developers.google.com/web/updates/2013/09/dialog-element-Modals-made-easy).

Using the `cancel` event should solve this. In your case, try with `cancel` instead of `onCancel`.

I'm thinking of cancelling the escape key by default, unless the user provide a `cancel` event. What do you think @Permagate?

Permagate:
Hi @tleunen, thanks for the link. there are some new things regarding the native dialog element that I have learned.

> Using the cancel event should solve this. In your case, try with cancel instead of onCancel.

I couldn't find the doc that says `onCancel` would be implemented, cmiiw. So I'm thinking if we should provide one ourselves? So that we don't need to attach a listener everytime we create a dialog, like @t1bb4r did.

> I'm thinking of cancelling the escape key by default, unless the user provide a cancel event.

Yeah, I agree.

tleunen:
The thing is, `cancel` is an attribute on the native dialog, so we don't need to attach an event.

tleunen:
I'll open a ticket in React to know if they plan to add the `onCancel` event.

tleunen:
[...]

PR #237, opened 2016-02-24
fix dialog cancel handler, update tests, and update documentation

Permagate:
Related to #221 as per latest discussion. I also updated the docs since to make the onCancel availability more obvious.

Regarding tests, I cannot exactly test the 'open' state as of right now. Since the native `open`, `showModal`, and the `cancel` event fired on pressing Escape can only be called when the dialog is mounted on real document. I only got element not in document error when I tried them. The very least I can do is to test if the listeners are attached and detached correctly.

Cheers!
Permagate.

### DIFF
--- a/docs/pages/dialogs.md
+++ b/docs/pages/dialogs.md
@@ -92,6 +92,51 @@ class Demo extends React.Component {
 }
 ```
 
+### Dialog with Cancel Event Handler
+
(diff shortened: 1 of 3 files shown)


tleunen:
I got the exact same issue when trying to write test around that.
Good enough, thanks so much!

PR #282, opened 2016-04-11
feat: Add prop to hide the spacer element in the header

drager:
This pull request adds an option to hide the spacer element in the header. I updated the docs as well with this change. Default is still to output the spacer.

Let me know what you think.


tleunen:
Thanks @drager for the contribution!

Issue #300, opened 2016-04-20
Build does not work on Windows


Download:
```sh
C:\ws\react-mdl>npm run build-umd

> react-mdl@1.5.3 build-umd C:\ws\react-mdl
> NODE_ENV=production webpack src/index.js out/ReactMDL.js --output-library ReactMDL --output-library-target umd

'NODE_ENV' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.10586
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\Frank\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "run" "build-umd"
npm ERR! node v5.4.0
npm ERR! npm  v3.5.3
npm ERR! code ELIFECYCLE
npm ERR! react-mdl@1.5.3 build-umd: `NODE_ENV=production webpack src/index.js out/ReactMDL.js --output-library ReactMDL --output-library-target umd`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the react-mdl@1.5.3 build-umd script 'NODE_ENV=production webpack src/index.js out/ReactMDL.js --output-library ReactMDL --output-library-target umd'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the react-mdl package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     NODE_ENV=production webpack src/index.js out/ReactMDL.js --output-library ReactMDL --output-library-target umd
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs react-mdl
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls react-mdl
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     C:\ws\react-mdl\npm-debug.log
```

Don't worry, I'm intending to fix this myself  :)

Issue #298, opened 2016-04-20
Dependency names wrong when using React MDL from script


Download:
I am including React MDL 1.5.3 from [npmcdn](https://npmcdn.com/react-mdl@1.5.3/out/ReactMDL.js).

If you follow that link and inspect the top of the file (UMD header) you see this line;

```js
root["ReactMDL"] = factory(root["react"], root["react-dom"]);
```
...which seems strange. I would expect `React` and `ReactDOM`.
And indeed, I am getting this error:

> ReactMDL.min.js:1026 Uncaught TypeError: Cannot read property 'Component' of undefined

When I inspect the `window` object, prior to React MDL loading, I see that the `React` object is there ,as expected, but not the `react` object that the UMD header is trying to load... again as expected.

Compare with ReactMDL.js at version [1.5.2](https://npmcdn.com/react-mdl@1.5.2/out/ReactMDL.js):

```js
root["ReactMDL"] = factory(root["React"], root["ReactDOM"]);
```

It seems something changed between 1.5.2 and 1.5.3 in Webpack's config that broke this.

Download:
I think it broke in this PR:
https://github.com/tleunen/react-mdl/pull/272/files

tleunen:
Thanks @Download for the investigation.
I'll revert the commit in https://github.com/tleunen/react-mdl/pull/299

PR #301, opened 2016-04-20
Issue 267

Download:
This PR addresses two issues:

* Build does not work on Windows #300
* Dependency names wrong when using React MDL from script #298

I've ran the build and the new UMD header looks like this:

```js
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory(require("react"), require("react-dom"));
	else if(typeof define === 'function' && define.amd)
		define(["react", "react-dom"], factory);
	else if(typeof exports === 'object')
		exports["ReactMDL"] = factory(require("react"), require("react-dom"));
	else
		root["ReactMDL"] = factory(root["React"], root["ReactDOM"]);
})(this, function(__WEBPACK_EXTERNAL_MODULE_2__, __WEBPACK_EXTERNAL_MODULE_4__) {
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "babel-preset-es2015": "^6.6.0",
     "babel-preset-react": "^6.5.0",
     "coveralls": "^2.11.9",
+    "cross-env": "^1.0.7",
     "eslint": "^2.7.0",
     "eslint-config-airbnb": "^6.2.0",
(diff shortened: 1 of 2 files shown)


tleunen:
Looks good @Download!
Thanks :)

Issue #302, opened 2016-04-21
data-upgraded causes Warning: React attempted to reuse markup in a container but the checksum was invalid


Download:
I'm using server-side rendering and have been getting warnings about React not re-using the server-side markup because it differs (ever so slightly) from the client generated markup. I'm now trying to get rid of these warning but I'm kinda stuck on this one:

> react.js:19287 Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
> (client) n" data-reactid="28">Login</button><nav 
> (server) n" data-reactid="28" data-upgraded=",Mat

I realize that `data-upgraded` is probably added by the material.js script which is actually not part of this project (or shouldn't be at least) and that it will be removed for v2, but I was just wondering if you have experience with this warning and whether you know/think it can be remedied or whether I should just live with it for now?

I have improved my server side and it's now pretty darn fast, but because the server markup can't be reused I get this ugly flash of unstyled content a second or so after the page has loaded.

Download:
My server does not actually render the `data-upgraded` attribute; it is added by material.js after it loads, but before the client has kicked in... Maybe there could be a way to prevent material.js from doing it's work until the client has actually kicked in and rendered for the first time?

Download:
I've done some more digging.

I think I could prevent the issue by changing this code in material.js:

```js
window.addEventListener('load', function() {
  'use strict';

  /**
   * Performs a "Cutting the mustard" test. If the browser supports the features
   * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
   * components requiring JavaScript.
   */
  if ('classList' in document.createElement('div') &&
      'querySelector' in document &&
      'addEventListener' in window && Array.prototype.forEach) {
    document.documentElement.classList.add('mdl-js');
    componentHandler.upgradeAllRegistered();
  } else {
    /**
     * Dummy function to avoid JS errors.
     */
    componentHandler.upgradeElement = function() {};
    /**
     * Dummy function to avoid JS errors.
     */
    componentHandler.register = function() {};
  }
});
```

My idea is to place this code in a separate function `initializeMDL`, and then let the script check a flag before it assigns this function to `onload`, like this:

```js
if (! window.suppressMDLonLoad) {
  window.addEventListener('load', initializeMDL);
}
```

[...]

PR #303, opened 2016-04-21
Allow to prevent auto-upgrade of components on load

Download:
Code now checks whether `window.mdlSuppressAutoUpgrade` is truthy and if so, skips auto upgrade.
SEE: data-upgraded causes Warning: React attempted to reuse markup in a container but the checksum was invalid #302 

This change allowed me to set a flag to prevent the auto-upgrade, which makes the issue go away.

Here is the surprise: I expected that I would have to call `upgradeAllRegistered` myself, but actually my app seems to be working just fine without that call. I suspect this is because React MDL already upgrades / downgrades all components when mounting / unmounting.  So probably, we can make this change even simpler and just remove the line that calls `upgradeAllRegistered` from the onload handler altogether. However to make this PR more acceptable, I opted not to do that and instead just use the flag. This will make sure that no one is affected by this change that does not specify this flag. 

We can always remove the flag later after more testing I think.

### DIFF
--- a/extra/material.js
+++ b/extra/material.js
@@ -58,8 +58,10 @@ var componentHandler = {
    */
   upgradeElements: function(elements) {},
   /**
-   * Upgrades all registered components found in the current DOM. This is
-   * automatically called on window load.
+   * Upgrades all registered components found in the current DOM.
+   * This is automatically called on window load, unless the flag
+   * `mdlSuppressAutoUpgrade` is present on `window` and has a truthy
+   * value, in which case you should invoke this function youself.
    */
   upgradeAllRegistered: function() {},
   /**
@@ -471,7 +473,6 @@ window['componentHandler'] = componentHandler;
 
 window.addEventListener('load', function() {
   'use strict';
-
   /**
    * Performs a "Cutting the mustard" test. If the browser supports the features
    * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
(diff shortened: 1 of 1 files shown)


Download:
@tleunen Can you let me know what needs changing in order to get this merged?

tleunen:
I'll investigate this weekend if we can remove the onload event. I believe we can, because as you said we already upgrade/downgrade everything on mount/unmount.

Download:
@tleunen 
Thanks! Yes I think that would be best actually, but I didn't want to go as far. But if we just remove it, we avoid having to define some ugly flag.

I'm not sure about removing the complete event handler BTW. It's basically just the call to `componentHandler.upgradeAllRegistered` I think. Because the rest of the onload handler is just checking the env for compliance and setting a `mdl-js` class on the body element. Since rendering directly into the body element is not recommended anyway, I don't think it will hurt React users. And having that env check around makes sense.

Download:
If you like, I can just modify this PR to avoid the flag and just skip the call to `componentHandler.upgradeAllRegistered`. Let me know.

tleunen:
Yes please update the PR to remove the upgrade call.
We cannot remove the class because some users could use it.

Thanks ;)

Download:
@tleunen 
> Yes please update the PR to remove the upgrade call.

Will do!

> Also, this also needs to be in the minified file.

Yes, I was looking at the build script to see how to generate it, but I could not find it.
If you could do that part it would be great.

tleunen:
Sure, no problem. Will manually merge the PR with the updated patch and the minified file

tleunen:
Manually merged. Thanks @Download

Issue #304, opened 2016-04-26
 [Textfield] Allow label be a component


Kuirak:
I'd like to use some other components (e.g.  `FormattedText` from react-intl) for the `label` on a text field. 

If i set the id to a value it works with the current code:
`<Textfield label={<FormattedText id="MyLabel"/>}  id="foo" />`

Maybe it could be changed that `customId` is generate by something like lodash `uniqueId` and the `propTypes` allow components as label.

tleunen:
The problem with a automatically generated id is that you cannot really guarantee that you id will be the same between the server rendering and the client rendering... That's why it was based on the label.

For the fact to allow custom component for the label, I don't have anything against it, except I'm not a big fan if this syntax. It doesn't seem very "React". Maybe I'm wrong.
One other thing could be to have `labelComponent` prop. It would give something like this:
```jsx
<Textfield labelComponent={FormattedText} label"MyLabel" id="foo" />
```
Might be a bit better. I should ask some other React devs to see what would be the best.

Kuirak:
Actually I like this syntax with passing a fully valid component as prop. It is basically the same syntax you use for the `<Header title={<Icon name="logo"/>} />`.

It helps knowing where the passed component will be used in the actual component.

In my opinion either label or labelComponent should be set but not both.

tleunen:
I'll go with the component inside `label` ;)

Chun-Yang:
Any one is working on this? I can give it a shot.

tleunen:
Not that I'm aware of ;-)

leifoolsen:
Is the generated id correct?

When I inspect the first textfield component in the [examples page](https://tleunen.github.io/react-mdl/components/textfield/) I see the following output.
```html
<div class="mdl-textfield mdl-js-textfield is-upgraded" data-upgraded=",MaterialTextfield" style="width: 200px;">
<input class="mdl-textfield__input" id="textfield-Text" name="">
<label class="mdl-textfield__label" for="textfield-Text">Text...</label>
</div>
```
The third textfield it has the same id-value. Would i not be better to generate a random id, e.g. id="textfield-Text-someuniquerandomcharsequence"? 

For screenreaders to work properly it is required to correlate label and input with a unique id.

tleunen:
Auto generating things is not that great because of server side rendering, you would need to have the same id on server than client to prevent a re-render.

I agree using the label as "id" is not great, but you can pass your own `id` to the Textfield and this id will be used instead.
[...]

PR #311, opened 2016-04-30
enable element in checkbox and textfield

Chun-Yang:
https://github.com/tleunen/react-mdl/issues/304

### DIFF
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -25,15 +25,17 @@ class Checkbox extends React.Component {
     }
 
     labelElement() {
-      const { label } = this.props
+        const { label } = this.props;
 
-      if (!label) {
-        return null
-      } else if (typeof label === 'string') {
-        return <span className="mdl-checkbox__label">{label}</span>
-      } else if (typeof label === 'object') {
-        return label
-      }
+        if (!label) {
+            return null;
+        } else if (typeof label === 'string') {
+            return <span className="mdl-checkbox__label">{label}</span>;
+        } else if (typeof label === 'object') {
+            return label;
+        }
+
+        return null;
     }
 
     render() {
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -24,14 +24,6 @@ class Checkbox extends React.Component {
         }
     }
 
-    formatLabel(label) {
-        if (typeof label === 'string') {
-            return <span className="mdl-checkbox__label">{label}</span>;
-        }
-
-        return label;
-    }
-
     render() {
         const { className, label, ripple, ...inputProps } = this.props;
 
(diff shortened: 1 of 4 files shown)


src/Checkbox.js
```diff
@@ -38,7 +52,7 @@ class Checkbox extends React.Component {
                     className="mdl-checkbox__input"
                     { ...inputProps }
                 />
-                {label && <span className="mdl-checkbox__label">{label}</span>}
+                {this.labelElement()}
```
tleunen: I'd keep the old syntax but use something like `{label && this.formatLabel(label)}`

With `formatLabel` returing the label inside the mdl element if it's a string, or just returning `label` for the others cases. It will simplify the function also since checking the object is not needed.

src/Textfield.js
```diff
@@ -60,13 +60,32 @@ class Textfield extends React.Component {
...
+            return (
+                <label className="mdl-textfield__label" htmlFor={customId}>
+                    {label}
+                </label>
+            );
+        } else if (typeof label === 'object') {
+            return label;
+        }
+
+        return null;
```
tleunen: Impossible to reach because `label` is required.

So you can just return `label` if it's not a string

src/Textfield.js
```diff
@@ -60,13 +60,32 @@ class Textfield extends React.Component {
...
+    }
+
     render() {
         const { className, inputClassName, id,
               error, expandable, expandableIcon,
               floatingLabel, label, maxRows,
               rows, style, ...otherProps } = this.props;
 
         const hasRows = !!rows;
+        if (!id && typeof label === 'object') {
```
tleunen: I'd prefer having this check in a custom proptype checker function. You can see an example here in the doc: https://facebook.github.io/react/docs/reusable-components.html

tleunen:
Thank you for the contribution @Chun-Yang 👍

Chun-Yang:
@tleunen Thank you for this well maintained package! Let me know if this is what we want.

tleunen:
From a code point of view, it looks good.

But i'd like to test it visually to see if it's ok. My main concern with using a specific element for the label is that it won't get the MDL classname, and because of that, the UI could be quite bad.

Chun-Yang:
I used link in checkbox label. It works just fine since the class name is preserved. I am not sure about the label for Textfiled though.

Chun-Yang:
How about this?

tleunen:
Yep it might be better, indeed :)
In most cases, this should be fine since it will be links (your case) or component used for i18n (like FormattedText, which is just text in the end).

szymonmichalak:
Thanks for adding that! I also have a use case for that in order to add label with asterisk (*) to identify which field is required, the same way as Material Design guidelines suggest: https://material.google.com/components/text-fields.html#text-fields-required-fields (with color change on focus).

Is there any chance that this PR will be merged soon?

vyerneni17:
Thanks for this, I was breaking my brain to get an HTML element inside the label prop. Cheers!

PR #318, opened 2016-05-20
if child element has id, Tooltip use this

tetsuyainfra:
This patch is a child element's id will be used for Tooltip component("for" attribute)

(I think you might know this already)
Background & Problem:
When server-side rendering, React will calculate the checksum from the elements .
if Tooltip components randomly generate id, redrawing will occur always on the client side.

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -4,7 +4,7 @@ import MDLComponent from './utils/MDLComponent';
 
 const Tooltip = (props) => {
     const { label, large, children, position, ...otherProps } = props;
-    const id = Math.random().toString(36).substr(2);
+    let id = Math.random().toString(36).substr(2);
 
     const newLabel = (typeof label === 'string')
         ? <span>{label}</span>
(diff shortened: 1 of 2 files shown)


tleunen:
Thanks @tetsuyainfra. For some reasons I missed this PR and I'm very sorry about that :/

With this change, I believe both the tooltip and the element will get exactly the same id, and we can't have twice the same id in the same page at the same time :/

tetsuyainfra:
Thanks @tleunen and your work, I know the id not should be used at the same time

This patch is when a child element have the id, tooltip use this id.
( if child element haven't id, the id is generated randomly )

However,  this patch is not a smart way, so programmer have to write id for ALL tooltip element on server-side rendering.

I will let you know if come up with workaround.

tleunen:
Yes I totally agree using a random id is not optimal at all.
I believe the best way would be to move the element as a child of the tooltip. That way we know the element without using any id. But it would also require a change in MDL I believe... They're working on their V2 so maybe we can suggest this change if it's not already planned.

see https://github.com/google/material-design-lite/issues/4495

Issue #326, opened 2016-06-14
Make Textfield render children


Download:
Currently, the Textfield component does not render it's children. I've found that that gives me issues sometimes where I really want my element (Icon etc) to appear inside the `<div class="mdl-textfield">...</div>` instead of next to it.

Changing this seems easy. Would you be open for a PR?

PR #327, opened 2016-06-14
Render `children` as last element of the Textfield

Download:
SEE Make Textfield render children #326


tleunen:
Thanks @Download. I think it's safe to say it cannot break anything ;-)

Download:
Thanks for merging it!

PR #332, opened 2016-06-23
Update react and react-dom

drager:
This pull request updates react and react-dom to the latest versions.


tleunen:
Unless I'm missing something, 15.1.0 is already covered

drager:
Yeah, they are covered as dependencies in this project but it would be great if peerDependencies also matched them and are not RC versions.

tleunen:
But the current peerDep also covers 15.1.0.

Not sure to understand why it doesn't cover `15.2.0-rc.1` though :/ https://semver.npmjs.com/

drager:
Ah alright, I see. About the `15.2.0-rc.1` not being included might be because it in face is an Release Candidate and not a stable release. So if you want to cover that as well you can do this: `^15.1.0 || 15.2.0-rc.1`. Not sure if you want that?

drager:
@tleunen: Saw that you updated the dependencies yourself, then you can close this PR?

PR #342, opened 2016-07-04
add default col property to Cell

Undeadlol1:
This is a small but usefull change.
This solves 2 problems:
When people forget to specify col property in Cell there will be no error and application will be fine.
When people want full width cell instead of writing
`<Cell col={12}> lorem ipsum </Cell>`
they can just drop
`<Cell> lorem ipsum </Cell>`
and code will be more readable then ever.


Undeadlol1:
I have no idea why my previous commits are being added to this PR. But seems like only one file is being changed which is exactly what i need.

tleunen:
Thanks @Undeadlol1, make sure you're up to date with master ;)

tleunen:
The default value is actually 4 (when the css class is not set). Should we use the same default or set it as 12 (or 4 or 8 depending on the platform)?

Undeadlol1:
@tleunen how come is it 4?
`col={12} === .mdl-cell--12-col ===  { width: calc(100% - 16px) }`
This basically applies full width on every media screen size. I checked it, there is nothing wrong.

tleunen:
In their doc, here's what they say:
```
mdl-cell--N-col          Sets the column size for the cell to N          N is 1-12 inclusive, defaults to 4; optional on "inner" div elements
```
https://getmdl.io/components/index.html#layout-section/grid

Undeadlol1:
They probably ment it defaults to 4 on mobile (and to 8 on tablet)
Then again, i tested this change and it works fine, so there must be a issue with documentation, not with the code.

src/Grid.js
```diff
@@ -31,16 +31,16 @@ Grid.propTypes = {
...
-    return typeof data !== 'undefined';
-}
-
 /* eslint-disable react/no-multi-comp */
 const Cell = (props) => {
     const { align, className, children, col, phone, tablet, component,
         hideDesktop, hidePhone, hideTablet, shadow, offset, offsetDesktop,
         offsetTablet, offsetPhone, ...otherProps } = props;
 
+    function isDefined(data) {
```
tleunen: Please revert this change

tleunen:
Gotcha. Could you squash the commits, make sure you're updated with the latest master and I'll merge ;)

Undeadlol1:
@tleunen Sorry for asking, but how do i squash commits? Any advices on how to understand hint better? Maybe you are using some GUI git client i don't know about?

Undeadlol1:
Git just seems a little frustrating to me...

tleunen:
Mo worries, I can do it if you have issues with it.
Squashing is done by doing an interactive git rebase `git rebase -i` (See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages)

tleunen:
I just committed manually the PR :) See f4e0a47fb964ade189fe002eec9c02bb970a494c

Thanks for your contribution @Undeadlol1

tleunen:
@Undeadlol1 I will actually revert the change, because the default value is 4, not 12. See this: https://github.com/tleunen/react-mdl/blob/master/extra/material-grid.css#L790

Ok the default depends on the viewport size, but still, it's not 12. I will remove the requirement of the prop, but I will not set any default.

Undeadlol1:
@tleunen you are confused.
With this commit i am trying to introduce logical behaviour to Cell component and improve readability of code.
There is simply no point in making this
`<Cell></Cell> === <div class="mdl-cell"></div>`
instead of this
`<Cell></Cell> === <div class="mdl-cell mdl-cell--12-col"></div>`
First option gives me 33% of page width, second one gives me 100%.
It's simply illogical. Why the hell do i need 1/3 width by default? What kind of random number is this? I don't even use 1/3 width components often, but i use full width all the time.

tleunen:
I agree, it might be weird, but for some reasons, that's the default value the MDL team came up to.
I would suggest you create a ticket under the [MDL project](https://github.com/google/material-design-lite) to potentially change the default width to 100% (equal to col=12).

There's no reason why we would change the default behavior of a cell in react-mdl compared to mdl. It might confuse the users if we don't use the same defaults.

Undeadlol1:
@tleunen I can see your point.
But i don't think anybody would be confused, because 99% of the time people read component's property-values-defaults table.
Also, i do not think they will change it, because seems like it's a workaround to make cell render properly on mobile devices. But i will try to add a PR/ create a issue.

tleunen:
Please create a ticket upstream, just so we can see their thoughts on it.
Anyway, this would be a breaking change so we can't really change this now

Undeadlol1:
I am not a smart man. What does "create ticket upstream" mean?
My intention were to make an issue at mdl with question about default width and link to this discussion. Afterwards i were planning to post link with mdl's answears here

tleunen:
Yes, it's a good idea to do that :) https://github.com/google/material-design-lite/issues

PR #343, opened 2016-07-05
fix(Snackbar): Prevent unknown prop warning

elasim:
delete onTimeout and timeout from otherProps to prevent warning.


tleunen:
Perfect, thanks @elasim :)

PR #353, opened 2016-07-19
Improve information for getting started

### LINKED ISSUE #352: Getting Started Documentation
When I was getting started with this project I had no idea how to include and access the library.
I ended up creating my own browser bundle with browserify, which worked for a while until I [ran into a problem](https://github.com/tleunen/react-mdl/issues/335) because my bundle contained a second copy of React.
I only found out from [this closed issue](https://github.com/tleunen/react-mdl/issues/239) that there is already a browser bundle, `ReactMDL.js`. It's just only on `npm` and not in the repo, even though there's [lots of other stuff](https://github.com/tleunen/react-mdl/tree/master/extra) in the repo for distribution purposes.
(I've since [updated my usage](https://github.com/1j01/natter/commit/5dd2828ce25705fcd3ab7f6cedeb6ff183564979), getting rid of webpack.)

The README should explain how to install and include the library.
You don't need to `import` it (which isn't supported in major browsers), or use browserify or webpack, you can just copy `node_modules/react-mdl/out/ReactMDL.js` to a `lib` or `vendor` folder and include it as a script (unless you have a setup where you can rely on the existence of `node_modules`)

I could create a PR with some improvements.

1j01:
Resolves #352

Didn't end up making specifically a "Getting Started" section, but yeah. Something like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,14 @@ If you're worried about the changes we made, you can take a look at the [patch n
 <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
 ```
 
-## How to use the components
+## Installation
+
+`npm install --save react-mdl`
+
+
+## Usage
+
+You can use React-MDL through a bundler like [webpack][webpack-url] or [browserify][browserify-url] by importing `react-mdl` in your application:
 
 ```js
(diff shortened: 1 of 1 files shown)


tleunen:
Looks way better than what we had before ;) Thanks @1j01

PR #363, opened 2016-08-12
Including docs on Tab parameters

### LINKED ISSUE #355: What's the best way to integrate Tabs with react-router?
I would like to use the MDL Tabs for my application that uses `react-router` for navigation...

Initially I was using the pure JS, but had some trouble with its events. Then, I found this library, but the best I could do still seems quite _ugly_:

```jsx
export default class NavigationBar extends Component {

  constructor(props) {
    super(props)
    this.state = {
      active: 0
    }
  }

  //FIXME: holy cow there MUST be a better way to use Tabs for navigation!
  onChange(id) {
    let node = findDOMNode(this)
    let path = node.children[0].children[id].attributes.to.nodeValue
    this.setState({ active: id })
    hashHistory.push(path)
  }

  render() {
    return (
      <Tabs ripple activeTab={this.state.active} onChange={this.onChange.bind(this)}>
        <Tab to="/">Home</Tab>
        <Tab to="/stuff">Stuff</Tab>
        <Tab to="/etc">Etc</Tab>
      </Tabs>
    )
  }

}
```

My first thought was something like `<Tab><Link to="/">Home</Link></Tab>` but that triggers an error because there is an anchor inside another :(

igorsantos07:
With this present we can close #355, as there's a feature that solves that issue but was not documented.

### DIFF
--- a/components/tabs/index.html
+++ b/components/tabs/index.html
@@ -137,6 +137,34 @@ var Demo = function (_React$Component) {
 <td style="text-align:left">Pass Props to the TabBar Element</td>
 <td style="text-align:left">Optional</td>
 </tr>
+<tr>
+<td style="text-align:left">Tab</td>
+<td style="text-align:left">active</td>
+<td style="text-align:left">bool</td>
+<td style="text-align:left">If this is the active tab</td>
+<td style="text-align:left">Optional</td>
+</tr>
+<tr>
+<td style="text-align:left">Tab</td>
+<td style="text-align:left">component</td>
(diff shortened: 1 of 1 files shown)


tleunen:
Thank you @igorsantos07! But it's against the wrong branch ;)

Could you create a new PR against master and update this file instead?
https://github.com/tleunen/react-mdl/blob/master/docs/pages/components/tabs.md

igorsantos07:
Whoops. Haha will do

Igor Santos
-- Desenvolvedor Web
[enviado do meu celular]

Issue #359, opened 2016-08-09
Support for immutable-js objects


ahoereth:
(Most?) react-mdl components, e.g. DataTable, only support js native arrays and objects as props. React itself also allows for direct use of immutable-js (and similar) datatypes since 0.13:

> [Support for iterators and immutable-js sequences as children.](https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html#new-features)

Changing this behavior currently is as easy as replacing [`PropTypes.arrayOf(...)`](https://github.com/tleunen/react-mdl/blob/e32f7f0/src/DataTable/Table.js#L18-L20) or similar with `PropTypes.any`. 

I am currently planning on solving this using a function custom type checker testing the objects constructor name. Any input on this? 

```javascript
PropTypes.oneOfType([
  PropTypes.arrayOf(PropTypes.object),
  function(props, propName, componentName) { [...] },
])
```

tleunen:
Great! How would you check if it's an iterator or immutable-js object?

PR #369, opened 2016-08-16
Allow iterables as prop types similar to react core.

ahoereth:
See issue #359.

Uses react's [getIteratorFn()](https://github.com/facebook/react/blob/v15.3.0/src/shared/utils/getIteratorFn.js) to check weather the passed prop can be iterated. Currently I do not check for valid child items.. Not sure whether that's required? Would cause quite a bit of overhead.

Side note: Does this make sense for any other react-mdl component? Did not find anything where immutable (or similar) props make sense.

### DIFF
--- a/src/DataTable/Selectable.js
+++ b/src/DataTable/Selectable.js
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import getIteratorFn from 'react/lib/getIteratorFn';
 import classNames from 'classnames';
 import isEqual from 'lodash.isequal';
 import TableHeader from './TableHeader';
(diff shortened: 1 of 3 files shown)


tleunen:
Thanks @ahoereth, I believe it's ok. Maybe you could add a quick test validating only an object or iterator can be passed? (check that no error is thrown when rendering)?

tleunen:
If the user pass an immutable-js list instead of a plain array, and we're using array functions on it, how will it work on the immutable-js structure? Does it has the same functions?

Like, `.lenth()`, `.slice()`, `.map`, etc?

ahoereth:
There is no length property, `slice()`, `map()` etc work the same. For all the available functions see the sidebar in the [immutablejs docs](https://facebook.github.io/immutable-js/docs/#/List) -- they mostly correspond to their JS/ES6 equivalents, especially to the new [ES6 Map type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map). 

...and I just figured: This makes a lot more changes necessary. Seems like the code just seemingly works fine as is, but actually does not. Will need to add immutablejs as dev dependency to test this thoroughly.

ahoereth:
I won't have time to complete this PR in a timely manner, using a [simplified custom DataTable](https://gist.github.com/ahoereth/d417db3403ef82c9ec0eee7b23b7a632) for now. Closing this for now, hope to come back to it later.

PR #373, opened 2016-08-17
Using window height instead of body.clientHeight

pablocubico:
Position change for Dialogs, using screen height instead of body height.


tleunen:
Thanks @pablocubico!

PR #376, opened 2016-08-20
fix: Fix js error when clicking on a tab

### LINKED ISSUE #347: Fixed tabs in demo are broken
Trying to to change between fixed tabs in the Layout section is broken, throws an exception. Also lot's of "React is not defined" error floating around 

screen shot 2016-07-13 at 11 51 14 am

tleunen:
Fixes #347 

MDL tabs require a `href` to be set, so we're using a small hack to make it work without MDL errors.


mr-e-:
I am still seeing this issue in the latest npm package 1.7.2. Is there any other workaround for this?

PR #379, opened 2016-08-21
Including docs on Tab parameters

igorsantos07:
Correct version of #363 (a bit late, sorry!)


tleunen:
I'm not sure we should specify all props on Tab, because most of them are added by the `TabBar`.
Maybe only `component` should be specified?

Issue #374, opened 2016-08-17
Refs triggering errors


pablocubico:
I'm getting a few errors on components using string refs, as **Dialog** and **Textfield**.

<img width="1411" alt="captura de pantalla 2016-08-17 a las 12 45 12" src="https://cloud.githubusercontent.com/assets/340766/17743882/d338246c-647b-11e6-9dd8-5c4d50cca2e0.png">

See: https://facebook.github.io/react/warnings/refs-must-have-owner.html (not very helpful)

I solved it by using callback refs instead of string refs. Example here:

https://github.com/tleunen/react-mdl/compare/master...pablocubico:using-callback-refs-instead-of-strings?expand=1

String refs are on its way to deprecation as mentioned in: 
- https://facebook.github.io/react/docs/more-about-refs.html
- https://github.com/facebook/react/issues/6250

@tleunen Would you agree on switching all string refs to callbacks?

tleunen:
Not sure to understand why this is failing with string refs, but I definitely agree to change all string refs for string callbacks ;)

Do you want to help on this? A PR is very welcome :)

pablocubico:
The problems of string refs have been outlined in an answer here by Dan Abramov, checkout specially the last two points (3 & 4):

https://news.ycombinator.com/item?id=12090484

I'll make a PR.

PR #382, opened 2016-08-22
Using callback refs instead of strings

pablocubico:
Changes string refs into callback refs.

Fixes: #374


tleunen:
Looks good, thanks @pablocubico !

PR #388, opened 2016-09-03
Add instructions for Create React App

gaearon:
Please see https://github.com/facebookincubator/create-react-app/issues/564 for the reference.
It is an increasingly popular way of creating apps so it’d be great to mention this incompatibility.


tleunen:
Thanks @gaearon!
Yep, importing the files from the javascript is probably the way users will use the most. Not specific to "create-react-app".

PR #395, opened 2016-09-10
docs(DataTable): changed DataTable by Table

gabrielperales:
I saw that the component DataTable is switched by Table so I have updated the documentation


tleunen:
Both actually exist, but because the MDL team is using the term `data-table`, I'd prefer keeping `DataTable` as the suggested name for now.

PR #396, opened 2016-09-10
Updated DataTable documentation

gabrielperales:


tleunen:
Thanks @gabrielperales but we don't need to manually update `gh-pages`. This is done automatically when a new version of the project is released.

PR #398, opened 2016-09-13
Add the possibility to style Table rows by providing a 'style' attribute

Augustin82:


src/DataTable/Table.js
```diff
@@ -65,7 +65,7 @@ class Table extends React.Component {
                 </thead>
                 <tbody>
                     {realRows.map((row, idx) =>
-                        <tr key={row[rowKeyColumn] || row.key || idx} className={row.className}>
+                        <tr key={row[rowKeyColumn] || row.key || idx} className={row.className} style={row.style}>
```
tleunen: Hmm...

I know we're using that for `row.className` but I wonder if with `style` it could produce errors. Because if the rows data have `style` as a key, and that style is a string for example, React will not be happy.

I wonder if we could find a better way to allow the row styling (even for the className to be honest).

Augustin82: So do I... ;)

However, in the meantime, since there is support for className, I believe it only makes sense to have support for style, even if it means handling the entire thing differently later on.

What checks on the nature of 'style' should be performed to make it safe?
What checks on 'className' ?
Also, maybe 'style' should be renamed to something less likely to collide, like 'rowStyle' ?

Augustin82:
@tleunen any update on how I should proceed? Do I need to write more tests, to fix the handling of edge cases, to argue my case some more?

Please let me know and I'll get to work. Cheers!

tleunen:
Sorry @Augustin82, I've been busy with some other things recently and couldn't spend much time on react-mdl :s 

So after a quick investigation (to remind me how the Table works), `className` is mostly used internally to set the row as "selected". While I believe it's ok, I also think it could conflict with the row data, especially if the row data already has a `className` or `style` property.
I believe for v1, we should be fine, but for v2, I'd prefer having a special prop, maybe something like `__mdlRowProp` or something very unlikely to be used as a data prop to avoid any conflict.

Thoughts?

Augustin82:
@tleunen no worries, I was just pinging you. Thanks for taking the time to answer.

I agree with what you said on the need for v2 to have a special prop that would help limiting conflicts (along with explicitly defining it in the documentation).

Would it be OK, for the current version, to release support for "proper" React styling, still? I could update my PR to use a less obvious name than style (such as `__mdlRowClassName` and `__mdlRowStyle`, or something like you suggested).

The same thing could be extended to individual cells if deemed necessary; again, I could update my PR to support this.

Thanks!

tleunen:
On a cell basis, it could be a bit more complicated to do so. To allow that, each cell data must be an object, with the special style key inside. Which will require users to create a new array of objects. Initially, the way I designed the table was especially to be able to take the raw data from an API/DB and use it inside the table with only the keys we want to display, without manipulating the data (or manipulating while we're displaying with the cellFormatter).

It really depends on the use case, if there's one important use case for that, we might support it, otherwise it doesn't look necessary. Most of the time, a user should be able to achieve what he wants with a custom cellFormatter.

I believe your implementation is fine for now, maybe just check `style` is an object before adding it in the element?
And maybe rename it with `_mdlRowProps`, while adding support for custom className and style? Having a generic `props` would also allow to add onClick event and other things on the `tr`

Augustin82:
OK, I'm on it. Thanks for the feedback and suggestions =)

Augustin82:
Updated. What do you think, @tleunen ?

src/DataTable/Table.js
```diff
@@ -21,6 +21,23 @@ const propTypes = {
     shadow: PropTypes.number
 };
 
+const isString = (s) => typeof(s) === 'string' || s instanceof String;
+
+const getCustomRowProps = (row) => {
```
tleunen: hmm.. Why do we really want to check the received data?
There's a high chance the user know what to do when he uses `mdlRowProps`... I mean.. I hope he knows what he's doing ;p

src/DataTable/Table.js
```diff
@@ -64,11 +81,20 @@ class Table extends React.Component {
...
-                    )}
+                    {realRows.map((row, idx) => {
+                        const customRowProps = getCustomRowProps(row);
+
+                        return (
+                            <tr
+                                key={row[rowKeyColumn] || row.key || idx}
+                                style={customRowProps.style}
+                                className={customRowProps.className}
+                                onClick={customRowProps.onClick}
```
tleunen: I believe we can just pass everything without checks... Something like: `{ ...(row.mdlRowProps || {}) }`

tleunen: Also, what about the old `row.className`?

src/DataTable/Table.js
```diff
@@ -64,11 +81,20 @@ class Table extends React.Component {
...
-                        </tr>
-                    )}
+                    {realRows.map((row, idx) => {
+                        const customRowProps = getCustomRowProps(row);
+
+                        return (
+                            <tr
+                                key={row[rowKeyColumn] || row.key || idx}
+                                style={customRowProps.style}
+                                className={customRowProps.className}
```
tleunen: Also, what about the old row.className?

src/DataTable/Table.js
Augustin82: I wanted to make sure that we wouldn't be causing weird out-of-our-control React errors by allowing the wrong types through to React props.

If you think this is needlessly complicated, I'll remove. Thanks!

src/DataTable/Table.js
Augustin82: I assumed we were replacing it. I'll merge instead.

src/DataTable/Table.js
tleunen: `mdlRowProps` would be the responsibility of the user

src/DataTable/Table.js
Augustin82: You're right, that's a pretty big miss. Sorry about that.

Which one would you favor? Adding to mdlRowProps (creating it in the process) or merging two separate objects?

I'd tend to favor keeping the current behaviour, only adding more classNames to it via mdlRowProps, in case people like me were already overloading row.className. What do you think?

tleunen: For now, merging them would be best. We'll rethink this later (hopefully for v2)

Augustin82:
Updated as per your comments. Thoughts?

Augustin82:
@tleunen sorry to ping you, any update on this?

src/DataTable/Table.js
```diff
@@ -64,11 +64,19 @@ class Table extends React.Component {
...
-                        </tr>
-                    )}
+                    {realRows.map((row, idx) => {
+                        const { style, className: mdlRowPropsClassName, onClick } = row.mdlRowProps || {};
+
+                        return (
+                            <tr
+                                key={row[rowKeyColumn] || row.key || idx}
+                                className={classNames(row.className, mdlRowPropsClassName)}
+                                {...{ style, onClick }}
```
tleunen: Why not passing all the other `mdlRowProps`, not just style and onClick?

tleunen:
Sorry I was away this weekend and came back yesterday, will make sure we merge this asap ;)

Augustin82:
Is there any problem with it?

Augustin82:
@tleunen any update on this, please?

tleunen:
I thought I asked why doing `{...{ style, onClick }}` instead of just passing all props `{ ...mdlRowProps }` because if the user wants another event on it, like `onDoubleClick` or `onMouseOver`, we'll also need to add them. We can't really maintain a whitelist imo.

Augustin82:
Apologies, I didn't understand it that way.

I have pushed an update, is that what you mean?

Thanks!

tleunen:
Thanks @Augustin82 and sorry again for the long long delay

Augustin82:
No need to apologize. Thank you very much for taking the time to review it, suggest correction, review it again because I messed up, and merging it ;)

PR #408, opened 2016-09-28
add onBlur to Textfield propTypes

quavmo:
Glad to update tests, docs if desired.


tleunen:
This is not really required because we're forwarding all unknown props to the input.
Do you have any reasons to specifically set the proptype?

quavmo:
I have no specific reason.

Issue #394, opened 2016-09-10
On 1.7.1 Tabs won't get upgraded automatically


igorsantos07:
Using `<MDLComponent><Tabs>[...]</Tabs></MDLComponent>` works :B

tleunen:
Are you talking about `Tabs` in the `HeaderTabs`, or the generic `Tabs` component?

Because [these tabs](https://tleunen.github.io/react-mdl/components/tabs/) work fine, I can see the ripple for example, same for the `HeaderTabs` in the example with the 6 tabs.

What's your issue exactly?

igorsantos07:
[These tabs](http://codepen.io/igorsantos07/pen/wzrobZ?editors=0010) do not display ripples :)
This codepen is based on the example's code - I just doubled the Tabs, so I could create one that was wrapped by MDLComponent to display my findings.

tleunen:
Hmm weird, it looks like it doesn't get upgraded. I'll investigate

igorsantos07:
As mentioned on #380, I'm bringing here the [codepen](http://codepen.io/igorsantos07/pen/pEgNrK?&editors=001) displaying HeaderTabs that seems to be upgraded (they do have the colored underline) but have no ripple effect

Place1:
Can confirm i'm getting this issue as well. Something i noticed was that it worked with the official MDL js/css. I'm porting a react UI that used MDL without this library. I had the Tabs all working and then my last step was to swap out the MDL css/js from google with the one from this library and then it broke.

not sure if that's any help but i thought i'd let you know :D

vitalets:
Also can confirm this issue. Would appreciate the fix..

tleunen:
Thanks for the info @Place1.
I'm very busy still for a few days so I won't be able to take a closer look and find a fix. So any help is welcome

gabrielbull:
Me too, having the same issue.

DirtyHairy:
The heart of this issue is the component upgrade not handling functional components correctly. In addition, ripple requires `upgradeElements` (not the plural :smirk:) instead of `upgradeElement` to be called on the element. Both issues never show up on the doc pages as `upgradeElements` is called on the whole page during bootstrap there.

The above PR fixes these issues.

tleunen:
Good catch @DirtyHairy! I'll take a look at the PR tomorrow.

DirtyHairy:
[...]

Issue #415, opened 2016-10-03
Unable to see ripple or underline on Tabs. Example codepen doesn't even work.


mr-e-:
https://tleunen.github.io/react-mdl/components/tabs/

On this link, while it works on this page, if you click the first example codepen, it does not work. There is no underline and no ripple. This is the same effect I got when I tried to place into my code.

hmaciasc:
I can confirm not even adding ripple fixes it

StrikeForceZero:
a dirty fix

```js
private componentDidMount() {
   window.componentHandler.upgradeAllRegistered();
}
```

DirtyHairy:
This is a duplicate of #394 that got fixed in 1.8.0 .

PR #421, opened 2016-10-20
Fix upgrade for functional components, upgrade Tabs recursively in order to fix ripple effect

DirtyHairy:
Hi all!

This PR fixes component upgrades to correctly handle stateless functional components.

In addition `Tabs` require a recursive update via `componentHandler.upgradeElements` in order for ripple to work. To this end, this PR introduces a recusive mode of operation in `MDLComponent`. Together, these two changes fix issues #394 and #415.

Both issues do not show up on the documentation pages. This is because the markdown bootstraps calls `upgradeElements` on the whole page. This PR removes this call, bringing documentation into line with the actual component behavior.

Cheers
-Christian

### DIFF
--- a/docs/pages/components/menu.md
+++ b/docs/pages/components/menu.md
@@ -32,7 +32,7 @@ title: Menu
 {/* Top left */}
 <div style={{position: 'relative'}}>
     <IconButton name="more_vert" id="demo-menu-top-left" />
-    <Menu target="demo-menu-top-left" valign="top">
+    <Menu target="demo-menu-top-left" valign="top" ripple>
         <MenuItem>Some Action</MenuItem>
         <MenuItem>Another Action</MenuItem>
         <MenuItem disabled>Disabled Action</MenuItem>
@@ -43,7 +43,7 @@ title: Menu
 {/* Top right */}
 <div style={{position: 'relative'}}>
     <IconButton name="more_vert" id="demo-menu-top-right" />
-    <Menu target="demo-menu-top-right" valign="top" align="right">
+    <Menu target="demo-menu-top-right" valign="top" align="right" ripple>
         <MenuItem>Some Action</MenuItem>
         <MenuItem>Another Action</MenuItem>
         <MenuItem disabled>Disabled Action</MenuItem>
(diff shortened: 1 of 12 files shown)


tleunen:
Good catch for the SFC upgrade. Initially, I let the component which required upgrade as class, and Tabs didn't need it. But this was before trying to fix it and then I forgot it was only for class component.

I'm glad you fixed it for the SFC. And good catch for the reason why it worked in the documentation, but not when users would use the component.

Thanks again for the investigation and the fix. Before merging, because of the change in the upgrade function, I just want to double check the code locally and see if everything is still ok.

tleunen:
Hi @DirtyHairy, I've tested the PR against the documentation and I have noticed the tabs and ripple inside the Layout header doesn't work properly. An error is thrown in the MDL Ripple about calling style and classlist on undefined.

But it looks fine for the Tabs component!

DirtyHairy:
Hi @Tommy! Yes, I noticed the same yesterday. That is a side effect of the upgrade now working - previously, nothing would happen without a global uprade, and now the upgrade works, but is incomplete. I have yet to find out what is going wrong here... A short term solution would be disabling the ripple effect for header tabs until we have a full fix.

DirtyHairy:
Hi @tleunen !

I have fixed the remaining issues and and included the fixand a bunch of ripple fixes for other components into the pull request. I cannot find any glitches in the documentation anymore now.

The issue that caused the errors in the console is a nasty race between MDL component handlers. If MDL is used on a static page, the Layout handler will run before the ripple handler on the Tabbar. The Layout handler tags the Tabbar for special treatment by the ripple handler by adding a class. However, in React, it seems that the Tabbar ripple handler is called before the Layout handler and, as the tag class is missing, attaches event handlers that will throw later.

I have modified `HeaderTabs` to explicitly add the tag class, which removes the race.

Cheers
-Christian

tleunen:
> I have modified HeaderTabs to explicitly add the tag class, which removes the race.

Could you explain this?
All the changes you made make sense now that the doc doesn't auto upgrade everything in the page, I'll double check just to make sure ;)

DirtyHairy:
> Could you explain this?

I am explicitly adding the class `mdl-js-ripple-effect--ignore-events` in JSX if ripple is enabled. Usually, this is dynamically set by the Layout handler before the Ripple handler executes but, due to the order in which react calls the lifecycle hooks, in our case Ripple executes first. Setting the class statically prevents the Ripple handler from erroneously attaching the failing event handlers.

So, I guess 'removing the race' is not a precise way to put what I did, 'evading the consequences' is more like it :smirk:

DirtyHairy:
Thanks for merging... :smile:

PR #422, opened 2016-10-22
Allow extra props on Badge component

HriBB:
Adds the ability to pass extra props such as `onClick` handler on `Badge` component. @tleunen please let me know if this makes sense.

### DIFF
--- a/src/Badge/index.js
+++ b/src/Badge/index.js
@@ -16,7 +16,7 @@ const propTypes = {
 };
 
 const Badge = props => {
-    const { children, className, text, overlap, noBackground } = props;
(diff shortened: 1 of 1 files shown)


tleunen:
I don't mind adding it for consistency, but in theory, because Badge only clones the child, you can pass the onClick on the child directly

HriBB:
It's because `react-portal` adds a click handler dynamically on target prop.
```
<Menu target={<Badge text="2" overlap><IconButton name={'person'}/></Badge>} align={'tr tr'}>
  <MenuItem/>
  <MenuItem/>
  <MenuItem/>
</Menu>
```

PR #435, opened 2016-11-21
Restore menu DOM after downgrade in order to allow react to cleanly unmount the component.

### LINKED ISSUE #426: Unmounting a Menu from the DOM throws an error
Check out the [testcase](https://codepen.io/dirtyhairy/pen/QKRPWw) on codepen. The pen initially displays a `Button` ("nuke"), an `IconButton` and an attached `Menu`. Clicking the button removes the `Button` and just displays icon button and menu, causing react to unmount the menu at child position 3 and remount it at child position 2. On clicking the button, the debugger shows that an exception is thrown in the process.

The reason for this bug is that MDL does not clean the DOM when the element is downgraded, leaving react with the wrapper container as the direct child of the surrounding `div` (instead of the original `ul`). React then tries to unmount the `ul` from its supposed parent (the `div`), causing the exception to be thrown. A possible solution is to perform the cleanup in `componentWillUnmount` after downgrading. I can prepare a corresponing pull request.

DirtyHairy:
This PR fixes #426

### DIFF
--- a/src/Menu.js
+++ b/src/Menu.js
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
+import { findDOMNode } from 'react-dom';
 import classNames from 'classnames';
-import mdlUpgrade from './utils/mdlUpgrade';
 import basicClassCreator from './utils/basicClassCreator';
 
 const propTypes = {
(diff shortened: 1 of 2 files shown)


DirtyHairy:
Yes, react-mdl 1.x is a nightmare when it comes to DOM maniupulations :smirk:

PR #442, opened 2016-12-09
fixed interaction between IconToggle and Tooltip

bradzacher:
The MDL Tooltip component uses the ID to find the element to attach the mouse events to.

IconToggle was putting the passed ID prop on the child, invisible input element, as opposed to the visible parent. This meant that you were not able to nest an IconToggle within a Tooltip.

This fix adds the ID to the parent instead, allowing the tooltip work.


tleunen:
Looks good, but the proptype is missing for this `id`. Could you add it please?

bradzacher:
@tleunen as requested. ID has been added to the proptypes

PR #457, opened 2017-01-19
Fix IE10 dataTable select not able to use HTML5 data attributes

awjreynolds:
Using react-mdl dataTable on IE10 noticed the select was failing as IE10 does not support data attributes. This ensures that it uses the 'old' way of getting attributes.


tleunen:
Thanks @awjreynolds

PR #472, opened 2017-02-23
Implementing password type for TextField.

saboya:
I did not write tests for this.


tleunen:
You can use `<Textfield type="password" />` to achieve the same thing.

PR #473, opened 2017-02-23
Adding maxlength property to TextField

saboya:
I did not write tests for this.


tleunen:
Not really needed because all the extra props will be passed to the input field. Thank you though.

PR #475, opened 2017-02-27
Detecting Dialog backdrop clicks.

saboya:
This PR allows the user to pass a function to be called when the Dialog backdrop is clicked by checking if the user click is inside the bounds of the modal dialog. It accounts for the possibility of fields that don't report clientY / clientX correctly (like selectfield).

This was tested with Google Chrome only.

### DIFF
--- a/src/Dialog/Dialog.js
+++ b/src/Dialog/Dialog.js
@@ -5,6 +5,7 @@ import classNames from 'classnames';
 const propTypes = {
     className: PropTypes.string,
     onCancel: PropTypes.func,
+    onBackdropClick: PropTypes.func,
     open: PropTypes.bool
 };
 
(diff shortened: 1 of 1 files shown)


src/Dialog/Dialog.js
```diff
@@ -14,6 +15,10 @@ const defaultProps = {
 
 class Dialog extends React.Component {
     componentDidMount() {
+        if (this.props.onBackdropClick) {
+            this.backdropClickCallback = this.onDialogClick.bind(this);
+            this.dialogRef.addEventListener('click', this.backdropClickCallback);
```
tleunen: Do not bind and add the event conditionally to a props. Always add the event, and inside the event, only do something if a prop is set.

saboya: Updated branch with requested change.

Issue #404, opened 2016-09-21
[DataTable] selectedRows as a prop


dymat:
Hey Folks!
I'm working with DataTable and a list of row-Objects on which I need to apply a filter.

Here is an example:

```
var rowsList = this.props.store.Persons
    .filter(
        person => (person.name.toLowerCase().indexOf(this.state.filter.toLowerCase()) != -1 ));
```

```
<Textfield
    label="Filter"
    onChange={(event) => {this.setState({filter: event.target.value})}}
/>

<DataTable 
    selectable 
    rows={rowsList} 
    rowKeyColumn="id"
    onSelectionChanged={(data) => console.log(data)}
> ... </DataTable>
```

When I apply a filter rowsList and rerender the Table-Component I loose my some IDs in the selectedRows-List. I need to persist these selections even if a filter is applied. 

So I propose to enable selectedRows as a prop which is used to have rows selected "by default".

```
<DataTable 
    selectable 
    rows={rowsList} 
    rowKeyColumn="id"
    onSelectionChanged={(data) => console.log(data)}
    selectedRows={[1,6,2,...]}
> ... </DataTable>
```

Or is there another way I can't see?

Greeting,
~dymat

tleunen:
Do you mean that if you select a row with the table unfiltered, then filter it in order to remove the selected row, and then change the filter again to see the row.. The row should still be selected?

If that's the case, maybe we could add a prop to check the value of something in the row instead of relying on an internal state.

So let's say you have your person with this:
```js
{
  name: 'X',
  lastname: 'Y',
  selected: true
}
```
Then we could add a new prop on the table, something like `selectedRowKey` (or maybe rowKeySelected to match the other rowKeyColumn :D)

dymat:
Wow! You are fast! Thanks for reply!

You got exactly my point. It would be great if you could add such a prop.

BTW: this new prop would be a good base for writing a pagination add on which keeps track of selected rows over all pages...

Thanks again!
Bye,
~dymat

tleunen:
For the pagination, it really depends on how it's built. I have a similar component at work and the selected elements are kept in memory, it doesn't care if there's an pagination or not ;)

PR #477, opened 2017-03-02
[DataTable] Selected rows as props

carlbarrdahl:
This PR allows you to pass an array of selected rows as a prop (`selectedRows`) as mentioned in #404.

```js
const rows = [
  { id: '1001', name: 'Row 1001' }, 
  { id: '1002', name: 'Row 1002' }
]

const selectedRows = ['1001']

<DataTable selectable rowKeyColumn="id" rows={rows} selectedRows={selectedRows}>
  <TableHeader name="id">Id</TableHeader>
  <TableHeader name="name">Name</TableHeader>
</DataTable>
```

### DIFF
--- a/src/DataTable/Selectable.js
+++ b/src/DataTable/Selectable.js
@@ -37,7 +37,7 @@ export default Component => {
             if (props.selectable) {
                 this.state = {
                     headerSelected: false,
-                    selectedRows: []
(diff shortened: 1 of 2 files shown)


tleunen:
Make sense. Thank you for the contribution @carlbarrdahl

PR #484, opened 2017-04-14
use prop-types package for PropTypes

adam187:
https://facebook.github.io/react/warnings/dont-call-proptypes.html

> Note: React.PropTypes is deprecated as of React v15.5. Please use the [prop-types library instead](https://github.com/reactjs/prop-types).


tleunen:
Thank you @adam187!

PR #488, opened 2017-05-16
Fix React import

samerbuna:
Importing * as React seems to trigger the React.PropTypes warning.

I don't see a reason for doing it either


tleunen:
Don't even know why it was * as React :) Thanks @samerbuna

